Issue #12, opened 2024-11-21
Refactor "withdrawFees" to allow anyone to call it


PaulRBerg:
Refactor the `withdrawFees` function to allow any account to call it, and always transfer the ETH fees to the contract admin. The `to` parameter is removed.

Rationale:

- Allowing any account to claim makes it possible to work with fee collector services such as [Mimic](https://x.com/mimicfi/status/1750917624832119150?s=46)
- The contract admin will be a multisig, which would make withdrawals cumbersome to perform

PR #13, opened 2024-11-21
refactor: remove caller constraint from`withdrawFees`

smol-ninja:
Closes https://github.com/sablier-labs/airdrops/issues/12


PaulRBerg:
Thank you!

I've realized that `collectFees` would be a better name instead of `withdrawFees` to make a distinction between the Sablier fees and the user-facing functions related to withdrawing streamed tokens.

PR #14, opened 2024-11-26
Polish tests and scripts

andreivladbrg:
Context: the PR https://github.com/sablier-labs/airdrops/pull/5 was merged without a full review, so i will leave some suggested changes here.

#### Changes

test: remove the missing return value token
test: remove unused defaults
test: remove unneeded assert function
chore: add SALT as immutable


smol-ninja (approved):
Looks good.

Issue #1, opened 2024-11-15
Account for `string` in Create parameters


smol-ninja:
Related https://github.com/sablier-labs/v2-core/issues/1086

smol-ninja:
@razgraf for the campaigns, what shape name should be used? When a user claims, it would make call to create functions. Should we use `""` i.e. blank or do we want to use a custom name (may be campaign name) or do we want to specify shape name when creating a campaign?

smol-ninja:
Sounds good.

razgraf:
> specify shape name when creating a campaign which means storing it as an immutable parameter and then use it in claim flow

This sounds good to me!

smol-ninja:
Question: Since `shape` is redundant in `MerkleInstant`, it makes sense to pass it as a separate parameter for Merkle LL and LT instead of `ConstructorParams`. But that would add lot more code the contracts that anticipated. Would it be OK to ignore `shape` in `MerkleInstant` for this release? This will let me include `shape` in `ConstructorParams`.

Note: In the future, a good approach would be to have a shared constructor parameter between `createMerkleLL` and `createMerkleLT` and a separate for `createMerkleInstant`. A lot of parameters are common between LL and LT but unfortunately due to the same `ConstructorParams` across all create functions, we cannot DRY'ify the code between LL and LT.

RFC @sablier-labs/solidity

PaulRBerg:
It should be fine to ignore `shape` in `MerkleInstant` in the next release.

I also agree that we should have separate constructor parameters for `createMerkleInstant`.

PR #18, opened 2024-11-26
feat: include shape in merkle contracts

smol-ninja:
Closes https://github.com/sablier-labs/airdrops/issues/1.

### Changelog:
- Generates `salt` by encoding `baseParams` instead encoding parameters separately. This helps to prevent "stack too deep error" caused after adding `shape` in `keccak256(abi.encodePacked(..))` of salt.
- Refactors `asset` to `token` as a result of https://github.com/sablier-labs/v2-core/pull/1097.
- Adds a variable `shape` of type `string`. Instead of reverting on longer than 32 chars length, it truncates the name to 32 chars.
- Reorders some test assertions alphabetically.
- Renames `NAME` to `CAMPAIGN_NAME` in `Defaults` contract.

@andreivladbrg the idea of storing shape as `bytes32` did not work because when the string of length less than 32 characters is stored as `bytes32`, it adds `0` to make it fit into 32 bytes slot which changes the length of the string upon conversion back to string. On the other hand, storing it as a string datatype preserves the string length as well in the storage. Therefore I decided to store it as a string and truncates its length if the length exceeds 32 characters.


PaulRBerg (changes requested):
feedback below

src/abstracts/SablierMerkleBase.sol
```diff
@@ -45,6 +45,9 @@ abstract contract SablierMerkleBase is
     /// @inheritdoc ISablierMerkleBase
     string public override ipfsCID;
 
+    /// @inheritdoc ISablierMerkleBase
+    string public override shape;
```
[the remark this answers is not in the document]
smol-ninja: `string` cannot have immutable property in Solidity. For it to be immutable, it must be a fixed sized type. I brainstormed a solution with @andreivladbrg for storing `shape` as `bytes32` which can be immutable. But the issue with that approach is included in the description of this PR.

src/abstracts/SablierMerkleBase.sol
```diff
@@ -69,6 +72,13 @@ abstract contract SablierMerkleBase is
         ipfsCID = params.ipfsCID;
         MERKLE_ROOT = params.merkleRoot;
         NAME = bytes32(abi.encodePacked(params.name));
+
+        // If the shape string exceeds 32 bytes, truncate it to prevent `claim` from reverting.
+        if (bytes(params.shape).length > 32) {
+            shape = string(abi.encodePacked(bytes32(abi.encodePacked(params.shape))));
+        } else {
+            shape = params.shape;
+        }
```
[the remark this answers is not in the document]
smol-ninja: I assume most shape names would be generated through the UI and will be smaller than 32 bytes. Given that shape is only for UI and is trivial for the protocol, reverting a tx would waster user's gas for something thats of little importance for the protocol functionality.

src/types/DataTypes.sol
```diff
@@ -12,13 +12,16 @@ library MerkleBase {
     /// @param ipfsCID The content identifier for indexing the contract on IPFS.
     /// @param merkleRoot The Merkle root of the claim data.
     /// @param name The name of the campaign.
+    /// @param shape The shape of Lockup stream useful to differentiate between streams in the UI. It is truncated if
```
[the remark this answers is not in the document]
smol-ninja: https://github.com/sablier-labs/airdrops/pull/18/commits/6aa7c01ab1577bc1dbb4b639e4b73f8b81566b8a

src/interfaces/ISablierMerkleBase.sol
```diff
@@ -53,6 +53,9 @@ interface ISablierMerkleBase is IAdminable {
     /// @notice Retrieves the name of the campaign.
     function name() external returns (string memory);
```
andreivladbrg, line 54: should we be more explicit with this? since we now have 2 strings, which both are considered "names"

src/abstracts/SablierMerkleBase.sol
```diff
@@ -69,6 +72,13 @@ abstract contract SablierMerkleBase is
         ipfsCID = params.ipfsCID;
         MERKLE_ROOT = params.merkleRoot;
         NAME = bytes32(abi.encodePacked(params.name));
+
+        // If the shape string exceeds 32 bytes, truncate it to prevent `claim` from reverting.
+        if (bytes(params.shape).length > 32) {
+            shape = string(abi.encodePacked(bytes32(abi.encodePacked(params.shape))));
+        } else {
+            shape = params.shape;
```
andreivladbrg, line 80: what if we use the same approach for  campaign name? store it as string and truncate it have only 32 bytes

we can add a helper function `truncateSToB32`

andreivladbrg (commented):
generally LGTM 👍

src/SablierMerkleFactory.sol
```diff
@@ -86,17 +86,7 @@ contract SablierMerkleFactory is
...
-                msg.sender,
-                baseParams.token,
-                baseParams.expiration,
-                baseParams.initialAdmin,
-                abi.encode(baseParams.ipfsCID),
-                baseParams.merkleRoot,
-                bytes32(abi.encodePacked(baseParams.name))
-            )
-        );
+        bytes32 salt = keccak256(abi.encodePacked(msg.sender, abi.encode(baseParams)));
```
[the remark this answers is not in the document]
smol-ninja: Haha. I had the same question.

src/abstracts/SablierMerkleBase.sol
smol-ninja: Sounds good to me.

src/interfaces/ISablierMerkleBase.sol
andreivladbrg: `shape` and `ipfsCID` are self explanatory, but IMO `name` can lead to confusion
but, i am ok to keep the current as i don't have a strong opinion

smol-ninja: I think I understood your point. Also, the DX is better if we use `campaignName` instead of `name`. I am in favour of it.

andreivladbrg: ok, let's use `campaignName`

andreivladbrg:
pushed a commit to address the feedback, lmk if it looks good @smol-ninja

smol-ninja:
Yes. Thanks for the commit. Looks good to me. Can you please approve the PR if there are no more suggestions?

PR #20, opened 2024-11-27
Polish tests

andreivladbrg:
#### Changes
- test: import Assertions from lockup 
- test: remove duplicated assertEq functions in Assertions
- test: update tests accordingly

### DIFF
--- a/tests/fork/merkle-campaign/MerkleLT.t.sol
+++ b/tests/fork/merkle-campaign/MerkleLT.t.sol
@@ -210,8 +210,7 @@ abstract contract MerkleLT_Fork_Test is Fork_Test {
             defaults.tranchesMerkleLT({
                 streamStartTime: defaults.STREAM_START_TIME_ZERO(),
                 totalAmount: vars.amounts[params.posBeforeSort]
-            }),
-            "tranches"
+            })
         );
         assertEq(lockup.getUnderlyingToken(vars.expectedStreamId), FORK_TOKEN, "token");
         assertEq(lockup.getWithdrawnAmount(vars.expectedStreamId), 0, "withdrawn amount");
--- a/tests/integration/concrete/lt/claim/claim.t.sol
+++ b/tests/integration/concrete/lt/claim/claim.t.sol
@@ -142,8 +142,7 @@ contract Claim_MerkleLT_Integration_Test is Claim_Integration_Test, MerkleLT_Int
         // It should create a stream with `STREAM_START_TIME` as start time.
         assertEq(
             lockup.getTranches(expectedStreamId),
-            defaults.tranchesMerkleLT({ streamStartTime: streamStartTime, totalAmount: defaults.CLAIM_AMOUNT() }),
-            "tranches"
+            defaults.tranchesMerkleLT({ streamStartTime: streamStartTime, totalAmount: defaults.CLAIM_AMOUNT() })
         );
         assertEq(lockup.getUnderlyingToken(expectedStreamId), dai, "token");
         assertEq(lockup.isCancelable(expectedStreamId), defaults.CANCELABLE(), "is cancelable");
--- a/tests/integration/concrete/lt/constructor.t.sol
+++ b/tests/integration/concrete/lt/constructor.t.sol
@@ -117,6 +117,6 @@ contract Constructor_MerkleLT_Integration_Test is Integration_Test {
 
         vars.actualTranchesWithPercentages = constructedLT.getTranchesWithPercentages();
         vars.expectedTranchesWithPercentages = defaults.tranchesWithPercentages();
-        assertEq(vars.actualTranchesWithPercentages, vars.expectedTranchesWithPercentages, "tranchesWithPercentages");
+        assertEq(vars.actualTranchesWithPercentages, vars.expectedTranchesWithPercentages);
     }
 }
--- a/tests/utils/Assertions.sol
+++ b/tests/utils/Assertions.sol
@@ -2,49 +2,12 @@
 // solhint-disable event-name-camelcase
 pragma solidity >=0.8.22;
 
-import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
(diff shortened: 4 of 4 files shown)


PaulRBerg:
Agree with Shub.

I will close this PR as we have to share the final commits for scoping the audit ASAP.

andreivladbrg:
to @smol-ninja 

> Why should this repo not have its own set of constants? What if we need to add a new constant in the future?

it is not about _not_ having its own set of constants; it's about not maintaining the _same code_ in both places, given they are the same. if the airdrops repo requires its own unique constants, i am more than happy to add them.

btw, i also think we should remove some constants that can be obtained from `forge-std`, as this: https://github.com/foundry-rs/forge-std/blob/83c5d212a01f8950727da4095cdfe2654baccb5b/src/StdUtils.sol#L23-L24

> Same concern about assertions. IMO this repo should not depend on lockup's utils contracts. It gives us less flexibility over having its own utils contracts

on this point, i disagree. what is the purpose of adding them in the NPM package if even we are not using the testing utils across our repos? with the same rationale, why don't we re-implement the `PRBMathAssertions` instead of importing and using them?

i am completely against keeping these:
https://github.com/sablier-labs/airdrops/blob/850245445c7d1fba9e347c1e931f5c4f3fcb305b/tests/utils/Assertions.sol#L20-L48

>See point 2 in https://github.com/sablier-labs/airdrops/pull/5#issue-2663552581, which imo makes sense. https://github.com/sablier-labs/airdrops/pull/5#issuecomment-2489256584 as well.

thanks for sharing this again. at the time i first read it, i wasn’t sure what you meant since i hadn’t reviewed the tests. now i understand.

>Why FEE is not more appropriate for Defaults? Fee is a contract variable that can be changed. 

actually, `FEE` for airdrops would be more appropriate as a constant, as it is declared as immutable in the merkle base contract. for lockup, it can be declared in defaults. wdyt?

---

to @PaulRBerg 

> I will close this PR as we have to share the final commits for scoping the audit ASAP.

i completely agree that we need the final commits for scoping the audit, but at the same time, given that tests are not included in the scope, i believe making changes to the tests is not relevant at all to the audit process

smol-ninja:
> it's about not maintaining the same code in both places

I believe the maintenance cost of `Constants` and `Assertions` is near zero. We rarely make changes to that. 

> if the airdrops repo requires its own unique constants, i am more than happy to add them

Hypothetically, its possible that in future, we only release a new version of lockup but no new version of airdrops. In that case, we may be interested into updating value of `JULY_1_2024` for lockups while keeping it same for airdrops. If let airdrops depend on lockup, we may need to think twice while making such changes.

> btw, i also think we should remove some constants that can be obtained from forge-std,

I remember discussing this sometimes in the past. The reason we chose to define our own variables is due to consistency with the names. Foundry has `UINT256_MAX` whereas we have `MAX_UINT256`.

> what is the purpose of adding them in the NPM package

Good point and I do not know. Wouldn't it be useful for integrators?

> why don't we re-implement the PRBMathAssertions instead of importing and using them

`Assertions` inherit from `PRBMathAssertions`. This is different. I am in favour of having Airdrops' `Assertions` inheriting from Lockup's `Assertions`.

>  am completely against keeping these:

OK. Then what do you think about above point?

> actually, FEE for airdrops would be more appropriate as a constant, as it is declared as immutable in the merkle base contract

But its not immutable in Factory where its also tested.

--------

I agree with @andreivladbrg there was no need to close this PR as it had no `src` related changes.

andreivladbrg:
> I believe the maintenance cost of `Constants` and `Assertions` is near zero. We rarely make changes to that.

true, but so was with murky's merkle contract, and we decided to import it

> Hypothetically, its possible that in future, we only release a new version of lockup but no new version of airdrops. In that case, we may be interested into updating value of JULY_1_2024 for lockups while keeping it same for airdrops. If let airdrops depend on lockup, we may need to think twice while making such changes

no problem at all with this, we simply won't update the airdrop repo with the new lockup version. we can have in `package.json` the NPM version in sync with airdrop. `@sablier/lockup": 2.0.0` or `github:sablier-labs/v2-core#commit_hash`. thus the only update made will be for lockup.

> I remember discussing this sometimes in the past. The reason we chose to define our own variables is due to consistency with the names. Foundry has UINT256_MAX whereas we have MAX_UINT256.

small thing for me, we can follow their way of declaring the constants (i.e. `UINT256_MAX`). and i actually prefer it, as it is easier alphabetically order them.

>  I am in favour of having Airdrops' Assertions inheriting from Lockup's Assertions.

great, in that case, we will need to remove the `assertEq` functions mentioned above.

<details><summary>to remove</summary>
<p>

https://github.com/sablier-labs/airdrops/blob/850245445c7d1fba9e347c1e931f5c4f3fcb305b/tests/utils/Assertions.sol#L20-L48

</p>
</details> 

> OK. Then what do you think about above point?

did you review the code? it was exactly what this PR was doing

> But its not immutable in Factory where its also tested.

fair enough, we can keep it in `defaults`

PaulRBerg:
I have reopened the PR, but I will let you debate the merits of it.

smol-ninja:
> small thing for me, we can follow their way of declaring the constants (i.e. UINT256_MAX). and i actually prefer it, as it is easier alphabetically order them

Fair point. Agree with you.

> did you review the code? it was exactly what this PR was doing

Yes I did. And I overlooked that this PR inherits Airdrops' `Assertions` from Lockup's `Assertions`. Very sorry about that. I saw that `Constants` got deleted and I wrongly assumed that `Assertions` got deleted as well.

> fair enough, we can keep it in defaults

Thanks. Lets do it. But since `FEE` is also declared in Lockup's `Constants`, we can't do it without renaming the variable. Other approach is to delete `FEE` from Lockup's `Constants` and define it in Lockup's `Defaults`. But this is what I was referring to. That having a `Constants` in airdrops would make it easier to make changes. Otherwise, to make it work, we would have to change it in Lockup, and then update commit in Airdrops' bun.

smol-ninja (approved):
Thanks. Looks good to me now.

PR #23, opened 2024-12-04
build: adds dependency from commit hash

smol-ninja:
Everytime there is a new commit in `v2-core`, the [CI fails](https://github.com/sablier-labs/airdrops/actions/runs/12152328558/job/33888533868) in this repo because the last commit of `v2-core` does not match with the commit hash from the bun lockfile. Therefore, this PR hard codes the commit hash of `v2-core`.


smol-ninja:
merging this otherwise CI would keep failing.

PR #24, opened 2024-12-09
test: check diff for admin balance in fork tests

smol-ninja:
This [fork CI failed](https://github.com/sablier-labs/airdrops/actions/runs/12218384157/job/34083834453) because foundry chose `users.admin` as one of the recipients. Thus, the following assertion failed:

```solidity
assertEq(users.admin.balance, fee, "admin ETH balance");
```

As a good practice, instead of comparing absolute balance, we should compare diff to avoid such scenarios (though highly unlikely). This PR fixes that.

Other than that, the failed CI did not send any notification on the Slack, so I figured out that the `notify-on-failure` has a different code than the one in [v2-core repo](https://github.com/sablier-labs/v2-core/blob/main/.github/workflows/ci-deep.yml#L78C1-L89C58). This PR fixed that as well.

### DIFF
--- a/tests/fork/merkle-campaign/MerkleInstant.t.sol
+++ b/tests/fork/merkle-campaign/MerkleInstant.t.sol
@@ -225,6 +225,6 @@ abstract contract MerkleInstant_Fork_Test is Fork_Test {
         merkleFactory.collectFees({ merkleBase: vars.merkleInstant });
 
         assertEq(address(vars.merkleInstant).balance, 0, "merkleInstant ETH balance");
-        assertEq(users.admin.balance - initialAdminBalance, fee, "admin ETH balance");
+        assertEq(users.admin.balance, initialAdminBalance + fee, "admin ETH balance");
     }
 }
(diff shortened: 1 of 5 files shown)


tests/fork/merkle-campaign/MerkleInstant.t.sol
```diff
@@ -223,6 +225,6 @@ abstract contract MerkleInstant_Fork_Test is Fork_Test {
         merkleFactory.collectFees({ merkleBase: vars.merkleInstant });
 
         assertEq(address(vars.merkleInstant).balance, 0, "merkleInstant ETH balance");
-        assertEq(users.admin.balance, fee, "admin ETH balance");
+        assertEq(users.admin.balance - initialAdminBalance, fee, "admin ETH balance");
```
andreivladbrg, line 228: wouldn't it be clearer if we: `initialAdminBalance  + fee == users.admin.balance`?

and similar below

andreivladbrg (approved):
LGTM, just 1 question:

PR #26, opened 2024-12-12
perf: store shape and campaign name in storage as bytes32

smol-ninja:
## Changelog
- Store campaign name and lockup shape as `bytes32` in the contract storage
- No change to the interface

## Motivation
@andreivladbrg noticed that the Factory contract size exceeded the limit even at just 50 runs. Upon further investigation, it became clear that storing values as strings was significantly increasing the bytecode size. By switching the storage variable type from `string` to `bytes32`, the contract size was reduced by 3,000.﻿ Its now 23,639 B which is well below the limit. 

## Trade-off
Since the String is first converted to `bytes32` and then back to the String during stream creation, it results in additional whitespace padding. For instance, "Airdrop Campaign" would ultimately be stored as `Airdrop Campaign                `. Most UIs, such as Etherscan's, are intelligent enough to trim these extra spaces.

One potential solution could involve storing the string length alongside the value, but this would introduce unnecessary complexity. In my opinion, this trade-off should be acceptable since the shape and campaign names are only intended for Sablier UI which can take care of it.

## Note
In the future, if needed, we can further reduce the size by 1000 bytes by using `bytes32` instead of `string` for the `campaignName` and `shape` fields in the input parameters of `ConstructorParams`.

### DIFF
--- a/src/abstracts/SablierMerkleBase.sol
+++ b/src/abstracts/SablierMerkleBase.sol
@@ -24,6 +24,9 @@ abstract contract SablierMerkleBase is
                                   STATE VARIABLES
     //////////////////////////////////////////////////////////////////////////*/
 
+    // The name of the campaign stored as bytes32.
+    bytes32 internal immutable CAMPAIGN_NAME;
+
     /// @inheritdoc ISablierMerkleBase
     uint40 public immutable override EXPIRATION;
 
@@ -36,17 +39,14 @@ abstract contract SablierMerkleBase is
     /// @inheritdoc ISablierMerkleBase
     bytes32 public immutable override MERKLE_ROOT;
 
+    // The shape of Lockup stream stored as bytes32.
+    bytes32 internal immutable SHAPE;
+
     /// @inheritdoc ISablierMerkleBase
     IERC20 public immutable override TOKEN;
 
     /// @inheritdoc ISablierMerkleBase
-    string public override campaignName;
-
-    /// @inheritdoc ISablierMerkleBase
     string public override ipfsCID;
-
-    /// @inheritdoc ISablierMerkleBase
-    string public override shape;
 
     /// @dev Packed booleans that record the history of claims.
     BitMaps.BitMap internal _claimedBitMap;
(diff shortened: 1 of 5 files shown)


src/abstracts/SablierMerkleBase.sol
```diff
@@ -36,18 +39,15 @@ abstract contract SablierMerkleBase is
     /// @inheritdoc ISablierMerkleBase
     bytes32 public immutable override MERKLE_ROOT;
 
-    /// @inheritdoc ISablierMerkleBase
-    IERC20 public immutable override TOKEN;
+    // The shape of Lockup stream stored as bytes32.
+    bytes32 internal immutable SHAPE;
```
andreivladbrg, line 43: let's use `@dev` tag

tests/utils/Defaults.sol
```diff
@@ -33,7 +33,8 @@ contract Defaults is Constants, Merkle {
     //////////////////////////////////////////////////////////////////////////*/
 
     uint256 public constant AGGREGATE_AMOUNT = CLAIM_AMOUNT * RECIPIENT_COUNT;
-    string public constant CAMPAIGN_NAME = "Airdrop Campaign";
+    // Since Factory stores campaign name as bytes32, extra spaces are padded to it.
+    string public constant CAMPAIGN_NAME = "Airdrop Campaign                ";
```
andreivladbrg, line 37: why not store it as `bytes32` also?

```solidity
bytes32 public constant CAMPAIGN_NAME_B32 = bytes32(abi.encodePacked(CAMPAIGN_NAME_S));
string public constant CAMPAIGN_NAME_S = "Airdrop Campaign";
```

andreivladbrg (commented):
It’s weird how much strings affect the size of the contract.
I’ve left some comments:

tests/utils/Defaults.sol
smol-ninja: Because:
1. Its explicit this way i.e. clearly shows the padded whitespaces. By using `CAMPAIGN_NAME_B32`, it would not be obvious as to why we have to do that. 
2. We compare this string against string returned by factory function so CAMPAIGN_NAME_B32 is redundant.

andreivladbrg: fair enough

PaulRBerg:
Wow I didn't know that strings are so expensive

PR #28, opened 2024-12-20
fix: use percantages in schedule

### LINKED ISSUE #27: Change the amount based variables in `MerkleLL.Schedule` to percentages
### Problem

When an airdrop campaign is created, it is very likely that not everyone will receive the same amount airdropped.

Which means we can have:
- A gets 100K
- B gets 10K
- C gets 10

The issue arises because the new unlock amounts variable stored in the `Schedule` struct is globally set for each stream.

https://github.com/sablier-labs/airdrops/blob/aa45dabb6c82c7cbf7ad02411e143bbe206f4a26/src/SablierMerkleLL.sol#L53-L54

https://github.com/sablier-labs/airdrops/blob/aa45dabb6c82c7cbf7ad02411e143bbe206f4a26/src/types/DataTypes.sol#L48-L54

This can result in situations where a creator wants to set a cliff amount of 1000 for A but only 1 for C, which is problematic. Furthermore, if the cliff amount exceeds the minimum amount airdropped, the stream cannot be created.

### Solution

The solution is to use percentage-based variables in the Schedule struct instead of amount-based ones, similar to how [TrancheWithPercentage](https://github.com/sablier-labs/airdrops/blob/aa45dabb6c82c7cbf7ad02411e143bbe206f4a26/src/types/DataTypes.sol#L64C12-L64C33) works.

---

Credits to @razgraf for this finding.

andreivladbrg:
Closes #27 

**Note:** unlike tranches percentages, here, we don't need to ensure that the unlock amounts sum equals the claim amount


src/SablierMerkleLL.sol
```diff
@@ -112,6 +112,11 @@ contract SablierMerkleLL is
             timestamps.end = timestamps.start + _schedule.totalDuration;
         }
 
+        // Calculate the unlock amounts based on the percentages.
+        LockupLinear.UnlockAmounts memory unlockAmounts;
+        unlockAmounts.start = ud60x18(amount).mul(_schedule.startPercentage.intoUD60x18()).intoUint128();
+        unlockAmounts.cliff = ud60x18(amount).mul(_schedule.cliffPercentage.intoUD60x18()).intoUint128();
```
andreivladbrg, line 118: since we can have the pct as 0, we could add `if(pct > 0) then mul`

but i decided not to do add this as it is not increasing the gas that much (0 mul is veerry cheap operation)
and it is shorter and clearer this one (just two lines ✨)

smol-ninja:
In SablierMerkleLL's constructor, why are you not verifying that the sum of percentages <= 100%? If it exceeds, campaign creator will not know until the first claim is made, which will then fail.

src/SablierMerkleLL.sol
smol-ninja: ternary operator ftw? :))

andreivladbrg:
> In SablierMerkleLL's constructor, why are you not verifying that the sum of percentages <= 100%? If it exceeds, campaign creator will not know until the first claim is made, which will then fail

due to this: https://github.com/cantinasec/review-sablier-3/issues/1
as we did for tranched pcts

src/SablierMerkleLL.sol
andreivladbrg: yeah, completely agree 😂

smol-ninja (approved):
Thanks for doing it so quickly. LGTM.

PR #39, opened 2025-01-06
fix: address audit findings 6, 8

smol-ninja:
This PR addresses the following audit findings:

- https://cantina.xyz/code/52d6aa1e-6aa4-4e73-96ed-6077652cf0f1/findings/6
- https://cantina.xyz/code/52d6aa1e-6aa4-4e73-96ed-6077652cf0f1/findings/8


src/abstracts/SablierMerkleBase.sol
```diff
@@ -59,10 +60,10 @@ abstract contract SablierMerkleBase is
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @notice Constructs the contract by initializing the immutable state variables.
-    constructor(MerkleBase.ConstructorParams memory params, uint256 fee) Adminable(params.initialAdmin) {
+    constructor(address campaignCreator, MerkleBase.ConstructorParams memory params) Adminable(params.initialAdmin) {
```
PaulRBerg, line 63: Can we please switch the order of these parameters so that `baseParams` comes first?

src/interfaces/ISablierMerkleFactory.sol
```diff
@@ -75,6 +72,10 @@ interface ISablierMerkleFactory is IAdminable {
                                  CONSTANT FUNCTIONS
     //////////////////////////////////////////////////////////////////////////*/
 
+    /// @notice Retrieves the fee for the provided campaign creator, using the default fee if no custom fee is set.
+    /// @param campaignCreator The address of the user for whom the fee is being computed.
+    function computeFee(address campaignCreator) external view returns (uint256);
```
PaulRBerg, line 77: I don't really like the name of this function.

Nothing is 'computed' here. It's just a variable that gets retrieved from storage depending upon a boolean flag.

Can we please rename it to `getFee`?

src/interfaces/ISablierMerkleFactory.sol
```diff
@@ -75,6 +72,10 @@ interface ISablierMerkleFactory is IAdminable {
...
+
     /// @notice Retrieves the custom fee struct for the provided campaign creator.
     /// @param campaignCreator The address of the user for whom the details are being retrieved.
     function customFee(address campaignCreator) external view returns (MerkleFactory.CustomFee memory);
```
PaulRBerg, line 81: And while we are at renaming fee-related functionality, could we also rename this getter to `getCustomFee`?

src/SablierMerkleFactory.sol
```diff
@@ -106,14 +111,11 @@ contract SablierMerkleFactory is
         // Hash the parameters to generate a salt.
         bytes32 salt = keccak256(abi.encodePacked(msg.sender, abi.encode(baseParams)));
 
-        // Compute the fee for the user.
-        uint256 fee = _computeFeeForUser(msg.sender);
-
         // Deploy the MerkleInstant contract with CREATE2.
-        merkleInstant = new SablierMerkleInstant{ salt: salt }(baseParams, fee);
+        merkleInstant = new SablierMerkleInstant{ salt: salt }(msg.sender, baseParams);
```
PaulRBerg, line 115: Can we use named parameters here

src/SablierMerkleFactory.sol
```diff
@@ -135,15 +137,12 @@ contract SablierMerkleFactory is
             abi.encodePacked(msg.sender, abi.encode(baseParams), lockup, cancelable, transferable, abi.encode(schedule))
         );
 
-        // Compute the fee for the user.
-        uint256 fee = _computeFeeForUser(msg.sender);
-
         // Deploy the MerkleLL contract with CREATE2.
-        merkleLL = new SablierMerkleLL{ salt: salt }(baseParams, lockup, cancelable, transferable, schedule, fee);
+        merkleLL = new SablierMerkleLL{ salt: salt }(msg.sender, baseParams, lockup, cancelable, transferable, schedule);
```
PaulRBerg, line 141: Named parameters

src/SablierMerkleFactory.sol
```diff
@@ -265,7 +254,7 @@ contract SablierMerkleFactory is
 
         // Deploy the MerkleLT contract with CREATE2.
         merkleLT = new SablierMerkleLT{ salt: salt }(
-            baseParams, lockup, cancelable, transferable, streamStartTime, tranchesWithPercentages, fee
+            msg.sender, baseParams, lockup, cancelable, transferable, streamStartTime, tranchesWithPercentages
```
PaulRBerg, line 257: Named parameters

tests/Base.t.sol
```diff
@@ -200,7 +199,7 @@ abstract contract Base_Test is Assertions, Constants, DeployOptimized, Modifiers
 
         bytes32 salt = keccak256(abi.encodePacked(caller, abi.encode(baseParams)));
         bytes32 creationBytecodeHash =
-            keccak256(getMerkleInstantBytecode(campaignOwner, token_, merkleRoot, expiration, fee));
+            keccak256(getMerkleInstantBytecode(caller, campaignOwner, token_, merkleRoot, expiration));
```
PaulRBerg, line 202: reorder this according to the order suggested in my other comments

PaulRBerg (changes requested):
Feedback below!

src/interfaces/ISablierMerkleFactory.sol
smol-ninja: Sounds good.

src/interfaces/ISablierMerkleFactory.sol
smol-ninja: OK.

src/SablierMerkleFactory.sol
smol-ninja: It will make it fit into multiple lines. Except for `msg.sender`, all the variables have the same name (baseParams will have params). Thats why I didn't use named parameters here but will do now.

PaulRBerg: I don't have an issue with multiple lines. Small cost to pay for extra clarity.

smol-ninja:
I've incorporated all your suggestion @PaulRBerg. LMK if it looks good now. 

A note that create merkle events no longer include fee since the fee logic has now been moved to the abstract contract. A solution to this could be to emit fee separately in the MerkleBase's constructor. Another solution is to recalculate fee in the create functions but that will add unnecessary gas for something that I believe is not so much important for the indexers. It can be fetched from the `FEE` variable anytime. LMK if thats important.

smol-ninja:
Pushed a new commit with the following changelog:

- Brings back the `fee` in CreateMerkle events.
- Introduces a private `_getFee` function for fee calculation in the Factory contract
- Replaces `uint64` with `uint256` in `isPercentagesSum100` function. Since its a function meant to be used only by the UI, it should return `false` for all other values except `1e18` instead of reverting.

src/SablierMerkleFactory.sol
```diff
@@ -61,25 +61,30 @@ contract SablierMerkleFactory is
                            USER-FACING CONSTANT FUNCTIONS
     //////////////////////////////////////////////////////////////////////////*/
 
+    /// @inheritdoc ISablierMerkleFactory
+    function getCustomFee(address campaignCreator) external view override returns (MerkleFactory.CustomFee memory) {
```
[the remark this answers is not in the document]
andreivladbrg: that's what i mean

remove `getCustomFee` and keep only `getFee`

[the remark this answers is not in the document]
smol-ninja: An argument to keep `getCustomFee` is that it is a getter to retrieve storage variable directly.

tests/Base.t.sol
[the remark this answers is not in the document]
andreivladbrg: the `caller` will always be the `campaignCreator`

src/SablierMerkleFactory.sol
```diff
@@ -172,27 +192,24 @@ contract SablierMerkleFactory is
             }
         }
 
-        // Compute the fee for the user.
-        uint256 fee = _computeFeeForUser(msg.sender);
-
         // Deploy the MerkleLT contract.
         merkleLT =
-            _deployMerkleLT(baseParams, lockup, cancelable, transferable, streamStartTime, tranchesWithPercentages, fee);
+            _deployMerkleLT(baseParams, lockup, cancelable, transferable, streamStartTime, tranchesWithPercentages);
```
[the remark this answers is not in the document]
smol-ninja: OK lets go with named args.

src/SablierMerkleFactory.sol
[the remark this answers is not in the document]
andreivladbrg: fair enough, let's keep it

andreivladbrg (approved):
shogun_lgtm

PaulRBerg:
The time is very close to my bed time and I am unable to review this tonight, so I will let @andreivladbrg handle this and hand off the final commits to the auditors.

> A note that create merkle events no longer include fee
> ...
> LMK if thats important.

That's a question for @sablier-labs/frontend, specifically ser @razgraf.

I'm slightly more inclined, though, to include the fee in the event as a just-in-case.

src/SablierMerkleFactory.sol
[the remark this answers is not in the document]
PaulRBerg: yeah let's definitely keep it; smol price to pay for knowing the custom fees

src/SablierMerkleFactory.sol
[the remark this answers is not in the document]
PaulRBerg: Named args are the way, guys. Always the way. When in doubt, go with named args.

andreivladbrg:
> I'm slightly more inclined, though, to include the fee in the event as a just-in-case

it is added:

https://github.com/sablier-labs/airdrops/blob/18652da1e1405f3e3d0d896527657dbf021458fc/src/SablierMerkleFactory.sol#L123

https://github.com/sablier-labs/airdrops/blob/18652da1e1405f3e3d0d896527657dbf021458fc/src/SablierMerkleFactory.sol#L166

https://github.com/sablier-labs/airdrops/blob/18652da1e1405f3e3d0d896527657dbf021458fc/src/SablierMerkleFactory.sol#L217

PR #48, opened 2025-01-27
build: update lockup dependency

smol-ninja:
Should be updated to lockup npm version.


andreivladbrg:
> Should be updated to lockup npm version

yeah, but the package is not yet published

should we merge it like this?

andreivladbrg (approved):

smol-ninja:
> should we merge it like this?

No, lets publish the lockup package first, then we can merge it and then we can publish the airdrops package.

PaulRBerg:
Agree with publishing the npm package first.

andreivladbrg (approved):

Issue #16, opened 2024-11-26
Separate out parameters of `createMerkleInstant` function


smol-ninja:
As discussed in https://github.com/sablier-labs/airdrops/issues/1#issuecomment-2501609932:

- [x] Add separate shared parameters struct for `createMerkleLL` and `createMerkleLT` functions. Then DRY'ify their internal logic.
- [x] Separate out the parameters of `createMerkleInstant` function and remove its dependency from the current `ConstructorParams` struct.

The objective is to separate out `createMerkleInstant` input parameters from `createMerkleLL` and `createMerkleLT`. Right now, there are a tons of parameters shared between `createMerkleLL` and `createMerkleLT` which can be DRY'ified through a common struct between them. However, due to them sharing it with `createMerkleInstant`, its not possible.

PR #54, opened 2025-01-31
refactor: merkle campaign constructors

smol-ninja:
### Changelog
- Closes https://github.com/sablier-labs/airdrops/issues/16

### DIFF
--- a/src/interfaces/ISablierMerkleFactory.sol
+++ b/src/interfaces/ISablierMerkleFactory.sol
@@ -4,7 +4,7 @@ pragma solidity >=0.8.22;
 import { IAdminable } from "@sablier/lockup/src/interfaces/IAdminable.sol";
 
 import { ISablierMerkleBase } from "../interfaces/ISablierMerkleBase.sol";
-import { MerkleFactory, MerkleInstant, MerkleLL, MerkleLockup, MerkleLT } from "../types/DataTypes.sol";
+import { MerkleFactory, MerkleInstant, MerkleLL, MerkleLT } from "../types/DataTypes.sol";
 import { ISablierMerkleInstant } from "./ISablierMerkleInstant.sol";
 import { ISablierMerkleLL } from "./ISablierMerkleLL.sol";
 import { ISablierMerkleLT } from "./ISablierMerkleLT.sol";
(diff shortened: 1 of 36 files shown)


andreivladbrg:
@smol-ninja since we decided to have separate factory for each merkle contract, is this PR still relevant?

smol-ninja:
Yes. We are only splitting once factory contract into four. That does not mean we will not have the shares data structure. IMO, lets merge this PR and then split the factory on top of the changes.

andreivladbrg:
but do we still need this: 

>> Changing `campaignName` and `shape` to `bytes32` has reduced the contract size by 3% (23,638 kB with 1000 runs).

smol-ninja:
> i believe having it as string is better for UX

Thats a good question. Lets ask @sablier-labs/frontend team if they are OK having `campaignName` as `bytes32`. I am fine with both., Except `ipfsCID`, we have other strings returning as `bytes32` as well. If the team is fine with the change, I am fine keeping it as `bytes32` as it does not offer a bad UX that much. The UI will still show the campaign name as string and on etherscan, I do not think the campaign name being `string` offers much value at all.

andreivladbrg:
well, why would we use `bytes32` if size is no longer an issue?
the `ipfsCID` discussion we had yesterday was only relevant to solution 3 from your discussion (proxy pattern). but now, since we will have 3 factories, it's no longer relevant.

smol-ninja:
> well, why would we use bytes32 if size is no longer an issue?

Fair point. I will change it to `string` but I would want to do it through a separate PR / issue (may be bundle it with factory splitting PR). https://github.com/sablier-labs/airdrops/pull/58 is built on top of this PR, so changing it would add a lot of work to that PR as well.

Is that OK?

andreivladbrg:
after checking the [roadmap](https://github.com/orgs/sablier-labs/projects/28/views/1) for the next airdrops iteration, this change will be the only *Hard Breaking Change* (when that issue was created we didn't have the [classification](https://github.com/orgs/sablier-labs/discussions/16)) 

thus, to make things easier for @sablier-labs/frontend team, maybe it would be better to leave them as they are (though there is no need for the `shape` parameter in Merkle Instant).

i will let you, @razgraf, decide whether we should move forward with this change or not.

razgraf:
As discussed on Slack, we're okay moving forward with this.

src/interfaces/ISablierMerkleFactory.sol
```diff
@@ -96,7 +111,11 @@ interface ISablierMerkleFactory is IAdminable {
     ///
     /// @param params Struct encapsulating the input parameters, which are documented in {DataTypes}.
```
smol-ninja, line 112: Natspecs are not updated. All look good otherwise.

smol-ninja:
@andreivladbrg can you please approve this PR if all looks good now?

andreivladbrg (approved):
> @andreivladbrg can you please approve this PR if all looks good now?

sure, i haven't checked the inherited components, but i trust you on this 



image

PR #55, opened 2025-02-01
test: use mainnet addresses in fork tests

smol-ninja:
### Changelog
- Change visibility of `_run` to `internal` and remove unneeded `virtual` from `run`
- Use mainnet address in fork tests
- Use actual Factory's admin in fork tests
- Add `--nmt testFork` in `test` and `test:lite` scripts. Use optimized profile to run fork tests locally

@andreivladbrg, it took half a day to figure out why the fork tests were failing locally. It is because the deployed contract is optimized so the default profiles cant make fork tests pass on deployed factory due to `CREATE2`.

### DIFF
--- a/tests/fork/Fork.t.sol
+++ b/tests/fork/Fork.t.sol
@@ -4,6 +4,7 @@ pragma solidity >=0.8.22 <0.9.0;
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { ISablierLockup } from "@sablier/lockup/src/interfaces/ISablierLockup.sol";
 import { Merkle } from "murky/src/Merkle.sol";
+import { ISablierMerkleFactory } from "src/interfaces/ISablierMerkleFactory.sol";
 
 import { Base_Test } from "../Base.t.sol";
 
@@ -29,12 +30,10 @@ abstract contract Fork_Test is Base_Test, Merkle {
 
     function setUp() public virtual override {
         // Fork Ethereum Mainnet at a specific block number.
-        vm.createSelectFork({ blockNumber: 21_745_237, urlOrAlias: "mainnet" });
+        vm.createSelectFork({ blockNumber: 21_719_244, urlOrAlias: "mainnet" });
 
-        // Set up the parent test contract.
-        Base_Test.setUp();
-
-        // Load the pre-deployed external dependencies.
+        // Load deployed addresses from Ethereum mainnet.
+        merkleFactory = ISablierMerkleFactory(0x71DD3Ca88E7564416E5C2E350090C12Bf8F6144a);
         lockup = ISablierLockup(0x7C01AA3783577E15fD7e272443D44B92d5b21056);
     }
 }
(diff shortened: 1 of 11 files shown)


tests/fork/Fork.t.sol
```diff
@@ -29,12 +31,17 @@ abstract contract Fork_Test is Base_Test, Merkle {
...
-        // Load the pre-deployed external dependencies.
+        // Load deployed addresses from Ethereum mainnet.
+        merkleFactory = ISablierMerkleFactory(0x71DD3Ca88E7564416E5C2E350090C12Bf8F6144a);
         lockup = ISablierLockup(0x7C01AA3783577E15fD7e272443D44B92d5b21056);
+
+        // Initialize the defaults contract.
+        defaults = new Defaults();
+
+        // Set the default fee for campaign.
+        resetPrank({ msgSender: merkleFactory.admin() });
```
andreivladbrg, line 44: should we declare here the admin:

```solidity
    users.admin = merkleFactory.admin();
    // or
    address internal factoryAdmin;
    function setUp() public virtual override {
           factoryAdmin = merkleFactory.admin();
     }
```

so that we don't declare it all merkle vars struct?

andreivladbrg (commented):
> it took half a day to figure out why the fork tests were failing locally. It is because the deployed contract is optimized so the default profiles cant make fork tests pass on deployed factory due to CREATE2

hmm, are you sure that this was the problem and not the fact that `Base.setUp()` was warping to a timestamp when the contracts didn't exist on the chain:

https://github.com/sablier-labs/airdrops/blob/76166a965f4736fc5210e8c46463febe082a1f88/tests/Base.t.sol#L95

smol-ninja:
> hmm, are you sure that this was the problem and not the fact that Base.setUp() was warping to a timestamp when the contracts didn't exist on the chain:

Yes I am 100% sure. If you run fork tests with `test-optimized` profile, they pass. I removed `Base.setUp()` from Fork test `setUp` as you can see in the diff.

tests/fork/Fork.t.sol
smol-ninja: good point. Will make the change.

smol-ninja: I will go with `factoryAdmin` variable for consistency since `users` is not used anywhere in the fork tests.

andreivladbrg:
> Yes I am 100% sure. If you run fork tests with `test-optimized` profile, they pass. I removed `Base.setUp()` from Fork test `setUp` as you can see in the diff.

🫡

tests/fork/Fork.t.sol
smol-ninja: https://github.com/sablier-labs/airdrops/pull/55/commits/2ffc44054b3a17aa4c08260993b8a72dc44f2cb6

smol-ninja:
If it looks good, can you please approve it @andreivladbrg? Thx.

andreivladbrg (approved):

Issue #43, opened 2025-01-10
Implement a new Merkle campaign with variable claim amount


smol-ninja:
As discussed in https://github.com/sablier-labs/company-discussions/discussions/85, implement a new merkle contract with the following features:
- The amount of airdropped token depends on the claim timestamp.
- Only works for ranged streams i.e. all streams begin at the same time.
- Enforce expiry to be greater than end time of the vesting schedule.
- Nomenclatures TBD.

```solidity
function createMerkleVCA(
    MerkleBase.ConstructorParams memory baseParams,
    MerkleLL.Schedule memory schedule,
    uint256 aggregateAmount,
    uint256 recipientCount
)
```

### Calculation of airdropped amount

```math
\text{airdrop amount} = \text{claim amount} \cdot \frac{\text{block timestamp} - \text{start time}}{\text{total duration}}
```

### Some questions
1. Should it have start and cliff unlocks?
2. What should the contract be called?

PaulRBerg:
Thanks for creating the issue.

> 1. Should it have start and cliff unlocks?

No. The first release should be as simple and slim as possible (to test the waters).

> 2. What should the contract be called?

Not an easy choice. Some ideas:

1. `MerkleLinear`
2. `MerkleVariable`
3. `MerkleVCA`
4. `MerkleFlex`
5. `MerkleAutoCancel`

I think the 1st would fit the best, but it might be confused with `LL` and/ or Lockup Linear streams.

smol-ninja:
I like `MerkleVCA` as its very clear once you know what VCA stands for.

1. `MerkleLinear`: As you said, can cause confusion with `LL`.
2. `MerkleVariable`: Sounds good, but "variable" can stand for many things.
3. `MerkleFlex`: Not speaking for itself.
4. `MerkleAutoCancel`: Technically, we are not cancelling any stream here. From the implementation's POV, the claim value is a function of time.

PaulRBerg:
Alright. `MerkleVCA` is the lucky winner.

andreivladbrg:
> function createMerkleVCA(
    MerkleBase.ConstructorParams memory baseParams,
    MerkleLL.Schedule memory schedule,
    uint256 aggregateAmount,
    uint256 recipientCount
)

why do we need a Schedule in AVCA? wasn't the plan to have only ranged streams?

smol-ninja:
[...]

PR #58, opened 2025-02-06
feat: `MerkleVCA` contract

smol-ninja:
### Changelog

- Closes https://github.com/sablier-labs/airdrops/issues/43

### PR dependency

- https://github.com/sablier-labs/airdrops/pull/54 must be merged first.


smol-ninja:
@andreivladbrg can you please review the source contracts? I'd like to have an agreement with you on them before writing tests.

smol-ninja:
This PR is now ready for your review @andreivladbrg.

src/SablierMerkleFactory.sol
```diff
@@ -186,6 +189,55 @@ contract SablierMerkleFactory is
...
+    function createMerkleVCA(
+        MerkleVCA.ConstructorParams memory params,
+        uint256 aggregateAmount,
+        uint256 recipientCount
+    )
+        external
+        returns (ISablierMerkleVCA merkleVCA)
+    {
+        // Check: neither vesting start time nor vesting end time is zero.
+        if (params.vesting.end == 0 || params.vesting.start == 0) {
```
andreivladbrg, line 202: I haven’t reviewed the full code yet, but I have a question regarding these checks.

Why not move them to the constructor? Since a `merkleVCA` can be deployed without using this factory specifically, placing the checks there would ensure that no contracts exist with incorrectly configured variables.

I know we are tracking only contracts deployed via our factory, but I believe it would be good practice to design `merkleVCA` in a way that prevents the possibility of misconfiguring variables.

smol-ninja: Good point. I thought about it and the reason I didn't place them in the constructor is because of the CEI pattern. The Merkle contract deployment is the "I" and these checks are the "C". So, reverting as soon as possible is what I went for.

Now, to your point that "placing these checks in constructor ensures that no contracts exist with incorrectly configured variables", I agree with you. It sounds like a good idea to me. What do you think about the CEI pattern, since its definitely gonna break it.

Also, to your point, given that these contracts will be shipped with BUSL, we cannot expect any other deployment anymore.

smol-ninja: I agree with your reasoning that no one should be able to deploy the contract with misconfigured variables but I do not agree that the pattern is respected. The EIC is not same as CEI.

smol-ninja: I've implemented your suggestion in https://github.com/sablier-labs/airdrops/pull/58/commits/e76d4e6b25a1e5e6fb98629c144290ba55b4d6ef. You will notice that I did not implement revert tests in the `contructor.t.sol` of merkle vca. This is because its already well tested in `createMerkleVCA.t.sol` of Factory.

andreivladbrg: > The EIC is not same as CEI

but do you agree that they are mutually dependent? if yes, it means they are both EIC and CEI at the same time (quantum stuff here :)) )

smol-ninja: Why not use the same argument throughout the codebase then and forget about CEI as long as all three are respected, in any order? I remember there are few places where CEI seems less intuitive but we still follow the pattern because we want to follow CEI pattern. Its by choice and also a good practice to follow.

andreivladbrg: > Why not use the same argument throughout the codebase then and forget about CEI as long as all three are respected, in any order? 

but where else in the codebase do we have this mutual inclusivity besides this?

smol-ninja: I am not sure what do you mean by "mutual inclusivity"? Can you explain with an example please?

smol-ninja: I see. So here is [a counter example](https://github.com/sablier-labs/lockup/pull/1069#discussion_r1831768143). 

By moving [`streamId = nextStreamId`](https://github.com/sablier-labs/lockup/blob/main/src/SablierLockup.sol#L425) into `function _create`, we can reduce the code by a few lines if we dont follow CEI. At the same time,  functions `_createLL` and `_create` are mutually exclusive. If any of the effect/interaction fails, the entire execution fails.

Its all Check, effect and interaction at the same time and there is no need to follow CEI in order.

andreivladbrg: oh, i didn’t understand what you were talking about before, but now i do.
the `SablierMerkleBase.constructor`, which runs first, vs `SablierMerkleVCA.constructor`, which runs second, has checks. you’re right, it’s not respected, but since the checks are specifically for `SablierMerkleVCA._vestingSchedule`, it’s a safe choice

src/libraries/Errors.sol
```diff
@@ -42,4 +42,20 @@ library Errors {
 
     /// @notice Thrown when trying to claim from an LT campaign with tranches' unlock percentages not adding up to 100%.
     error SablierMerkleLT_TotalPercentageNotOneHundred(uint64 totalPercentage);
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                 SABLIER-MERKLE-VCA
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @notice Thrown while claiming if vesting start time is in the future.
+    error SablierMerkleVCA_ClaimNotStarted();
```
andreivladbrg, line 51: let's say `SablierMerkleVCA_ClaimPeriodNotStarted(uint40 start, uint40 blockTimestamp)`

shell/prepare-artifacts.sh
```diff
@@ -24,12 +24,14 @@ cp out-optimized/SablierMerkleFactory.sol/SablierMerkleFactory.json $artifacts
...
 cp out-optimized/SablierMerkleLL.sol/SablierMerkleLL.json $artifacts
 cp out-optimized/SablierMerkleLT.sol/SablierMerkleLT.json $artifacts
+cp out-optimized/SablierMerkleVCA.sol/SablierMerkleVCA.json $artifacts
 
 interfaces=./artifacts/interfaces
 cp out-optimized/ISablierMerkleFactory.sol/ISablierMerkleFactory.json $interfaces
 cp out-optimized/ISablierMerkleInstant.sol/ISablierMerkleInstant.json $interfaces
 cp out-optimized/ISablierMerkleLL.sol/ISablierMerkleLL.json $interfaces
 cp out-optimized/ISablierMerkleLT.sol/ISablierMerkleLT.json $interfaces
+cp out-optimized/SablierMerkleVCA.sol/SablierMerkleVCA.json $interfaces
```
andreivladbrg, line 34: should be `ISablierMerkleVCA` 
missing "I"

src/SablierMerkleVCA.sol
```diff
@@ -0,0 +1,139 @@
...
+{
+    using SafeERC20 for IERC20;
+    using SafeCast for uint256;
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                  STATE VARIABLES
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @inheritdoc ISablierMerkleVCA
+    uint256 public override forgoneAmount;
```
andreivladbrg, line 45: there is a problem with this amount.

if the campaign is configured with `expiration == 0` (`clawback` is unusable), it means that any amounts left in the contract due to recipients claiming early would be lost.

to prevent this bug, we need to add a bespoke method to transfer the `forgoneAmount`:

```solidity
function transferForgoneAmount(address to) external onlyAdmin {
    // Check: the end time is in the past.
    if(endTime <= block.timestamp) {
        revert Errors.SablierMerkleVCA_ScheduleNotEnded();
    }

    // Interaction: transfer the forgone amount to the provided address.
    TOKEN.safeTransfer({ to: to, value: forgoneAmount });

    // Log the transfer.
    emit ForgoneAmountTransfer(admin, to, forgoneAmount);
}
```


--- 

this made me realize we should add tests for `when expiration == 0 then clawback is frozen`

<details><summary>something like this</summary>
<p>

```solidity
function test_WhenExpirationZero() external {
    ISablierMerkleBase campaignWithZeroExpiry = ISablierMerkleBase(createMerkleLT({ expiration: 0 }));
    assertFalse(campaignWithZeroExpiry.hasExpired(), "campaign expired");

    deal({ token: address(dai), to: address(campaignWithZeroExpiry), give: 1_000_000e18 });

    // make a claim to change `_firstClaimTime`
    campaignWithZeroExpiry.claim{ value: defaults.MINIMUM_FEE() }({
        index: 1,
        recipient: users.recipient1,
        amount: defaults.CLAIM_AMOUNT(),
        merkleProof: defaults.index1Proof()
    });

    // warp after the grace period
    vm.warp({ newTimestamp: block.timestamp + 10 days });

    resetPrank(users.campaignOwner);
    campaignWithZeroExpiry.clawback({ to: users.campaignOwner, amount: 100 });
}

```


</p>
</details>

src/SablierMerkleVCA.sol
```diff
@@ -0,0 +1,139 @@
...
+        uint128 claimableAmount;
+
+        // Calculate the claimable amount.
+        if (_vestingSchedule.end <= block.timestamp) {
+            // If the vesting period has ended, the recipient can claim the full amount.
+            claimableAmount = amount;
+        } else {
+            // Otherwise, calculate the claimable amount based on the elapsed time.
+            uint40 elapsedTime = uint40(block.timestamp) - _vestingSchedule.start;
+            uint40 totalDuration = _vestingSchedule.end - _vestingSchedule.start;
```
andreivladbrg, line 125: as we discussed on slack, i believe using PRBMath for extra precision is a good choice, especially for tokens with less than 18 decimals.

alternatively, we could do something similar to [solsab](https://github.com/sablier-labs/solsab/blob/6dd1a300283ed1da6a9288e3329d4a332bd800bd/programs/lockup/src/utils/streaming_math.rs#L23-L35).

andreivladbrg (changes requested):
while reviewing, a question came to mind—should we allow creating a `MerkleVCA` campaign with an end time in the past? or should we revert and instead, recommend the user to create a `MerkleInstant` campaign?

this would mean adding the following check in the constructor:

```solidity
if(endTime <= block.timestamp) {
   revert();
}
```

--- 

haven’t reviewed the tests yet, but in order to be more time-efficient, i’m going to leave these comments:

src/SablierMerkleVCA.sol
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /*//////////////////////////////////////////////////////////////////////////
+                          INTERNAL NON-CONSTANT FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @inheritdoc SablierMerkleBase
+    function _claim(uint256 index, address recipient, uint128 amount) internal override {
+        // Check: vesting start time is not in the future.
+        if (_vestingSchedule.start >= block.timestamp) {
```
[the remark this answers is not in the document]
andreivladbrg: 1a448b23c216104fcb54124b9c9ebcae3dc2ce46

shell/prepare-artifacts.sh
andreivladbrg: 1a448b23c216104fcb54124b9c9ebcae3dc2ce46

src/SablierMerkleFactory.sol
smol-ninja: I can say the same that the refactor in `_create` is also a safe choice (all tests would pass). Even though the CEI is not respected, I agree with you on this because the argument that someone else can deploy the VCA contract with misconfigured variables is very convincing.

src/SablierMerkleVCA.sol
smol-ninja: I do not agree with using PRBMath here. There is zero precision problem. The precision is absolute here. Token decimal is also irrelevant in this context since its a simple formula:

```math
\text{claim amount} = \frac{(\text{total amount} \times \text{elapsed time})}{\text{total time}}
```
In your `solsab` example, you are calculating `elapsed_time_percentage` first that is why it introduced precision problem.

```math
\text{claim amount} = \text{total amount} \times \frac{\text{elapsed time}}{\text{total time}}
```

You are performing division prior to multiplication. If you multiply first then divide, you lose 0 value and wont need scaling at all. Its not similar to Flow. Flow is different because of `rps`.

I may have been misunderstanding here but feel free to correct me in this case.

src/SablierMerkleVCA.sol
```diff
@@ -0,0 +1,139 @@
...
+        // Calculate the claimable amount.
+        if (_vestingSchedule.end <= block.timestamp) {
+            // If the vesting period has ended, the recipient can claim the full amount.
+            claimableAmount = amount;
+        } else {
+            // Otherwise, calculate the claimable amount based on the elapsed time.
+            uint40 elapsedTime = uint40(block.timestamp) - _vestingSchedule.start;
+            uint40 totalDuration = _vestingSchedule.end - _vestingSchedule.start;
+
+            claimableAmount = ((uint256(amount) * elapsedTime) / totalDuration).toUint128();
```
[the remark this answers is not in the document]
smol-ninja: The value `((uint256(amount) * elapsedTime) / totalDuration)` is of type `uint256` because of `uint256(amount)`, whereas `claimableAmount` is of type `uint128`.

src/SablierMerkleVCA.sol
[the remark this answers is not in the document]
smol-ninja: Good catch.

src/SablierMerkleVCA.sol
```diff
@@ -0,0 +1,139 @@
...
+        // Check: neither vesting start time nor vesting end time is zero.
+        if (params.vesting.end == 0 || params.vesting.start == 0) {
+            revert Errors.SablierMerkleVCA_VestingTimeZero({
+                startTime: params.vesting.start,
+                endTime: params.vesting.end
+            });
+        }
+
+        // Check: vesting end time is not less than the start time.
+        if (params.vesting.end < params.vesting.start) {
```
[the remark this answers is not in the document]
smol-ninja: Two reasons why I went with `<` instead of `<=`

1. `<` is more efficient than `<=`
2. Why shouldn't we allow start and end time to be the same time? I understand that it would become a classic example of instant airdrop, so nobody should use it for that. but having `<` is more gas efficient (even a tiny bit) over `<=`.

src/SablierMerkleVCA.sol
```diff
@@ -0,0 +1,139 @@
...
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                  STATE VARIABLES
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @inheritdoc ISablierMerkleVCA
+    uint256 public override forgoneAmount;
+
+    /// @dev The vesting schedule encapsulating the start time and end time of the airdrop unlocks.
+    MerkleVCA.Timestamps private _vestingSchedule;
```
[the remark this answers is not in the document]
smol-ninja: Yes. I absolutely agree. What do you think of calling it `_unlockTimestamp`. Simply `_timestamp` is fine too, but what do you think about prefixing it with "unlock"?

src/types/DataTypes.sol
```diff
@@ -123,3 +123,32 @@ library MerkleLT {
...
+        string ipfsCID;
+        bytes32 merkleRoot;
+        IERC20 token;
+        Timestamps vesting;
+    }
+
+    /// @notice Struct encapsulating the start time and end time of the airdrop unlocks.
+    /// @param start The timestamp when the airdrop token begins to unlock.
+    /// @param end The timestamp when the airdrop token unlocks 100%.
+    struct Timestamps {
```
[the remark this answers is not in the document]
smol-ninja: Because it only contains two timestamps. Where `Schedule` in `MerkleLL` contains durations and percentages. As per my https://github.com/sablier-labs/airdrops/pull/58#discussion_r1957084101, I think renaming the variable is a better idea.

src/libraries/Errors.sol
smol-ninja: Whats the benefit of including block time in the error message? It can easily be read from tons of places. I agree to include `start` in the error message though.

shell/prepare-artifacts.sh
smol-ninja: Sheesh. Copy Pasta failed. Thanks for catching it up.

src/SablierMerkleVCA.sol
smol-ninja: That was my first implementation and adding `transferForgoneAmount` has its problem too.

Imagine owner funding the campaign partially. Now all the forgone amount can be claimed by the late claimers because the campaign did not have enough funds. Or owner can also transfer forgone amount leaving late claimers not able to claim at all. Therefore, `endTime > block.timestamp` is a bad design choice. We still need a a time value which if `= endTime + 7 days` if we want to allow owner to transfer the forgone amount. So that after this time value has passed, late claimers cannot claim anymore and owner can transfer forgone amount to himself.

This value is similar to expiration only. Therefore, I then removed `transferForgoneAmount` from the contract.

------

An alternate solution is to not allow expiration to be 0 in this contract. Wdyt?

smol-ninja:
> while reviewing, a question came to mind—should we allow creating a MerkleVCA campaign with an end time in the past? or should we revert and instead, recommend the user to create a MerkleInstant campaign?

My reply is https://github.com/sablier-labs/airdrops/pull/58#discussion_r1957083642.

src/SablierMerkleVCA.sol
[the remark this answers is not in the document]
andreivladbrg: i don't think you get my point

i am simply saying that we can do this:

```solidity
claimableAmount =  uint128((uint256(amount) * elapsedTime) / totalDuration)
```

we don't need to use `SafeCast`

src/SablierMerkleVCA.sol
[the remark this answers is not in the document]
andreivladbrg: > but having < is more gas efficient

IMO the gas implications are irrelevant here since (i) a contract is being deployed and it is already an expensive operation (ii) campaign creators don't really care about the gas here (iii) as you said it is about a very tiny bit

i would rather chose a design that "makes sense" (instant unlock, use merkle instant - if there you want something else chose something else)  than this optimization

src/SablierMerkleVCA.sol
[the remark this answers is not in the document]
andreivladbrg: thanks.
i wouldn't use "unlock" keyword here. might be confused with the unlock amounts.
so i would chose simply `timestamps`

src/types/DataTypes.sol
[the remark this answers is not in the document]
andreivladbrg: agree, schedule can be confused with `MerkleLL.Schedule`

src/libraries/Errors.sol
andreivladbrg: i believe it is useful when debugging. in case of revert you have both returned.
but agree it is a small thing

src/SablierMerkleVCA.sol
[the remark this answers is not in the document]
smol-ninja: Ah I see. It was not clear in your comment. I agree with you.

src/SablierMerkleVCA.sol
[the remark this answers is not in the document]
smol-ninja: OK fair enough. I will introduce `<=` here.

src/SablierMerkleVCA.sol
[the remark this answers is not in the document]
smol-ninja: I don't think `unlockTimestamp` cannot be confused with `unlockAmount` since they end with distinct keywords. I am fine calling it `_timestamp`. I personally prefer to use bespoke names for variables and functions than using generic keywords such as `timestamps`.

src/libraries/Errors.sol
smol-ninja: Errors are only useful in a non-smart contract environment (similar to events). And in those environments, `block.timestamp` is very easy to read, as its always close to likely current timestamp plus tx timestamp is also available in these environments.

src/SablierMerkleVCA.sol
andreivladbrg: > Imagine owner funding the campaign partially. Now all the forgone amount can be claimed by the late claimers because the campaign did not have enough funds. Or owner can also transfer forgone amount leaving late claimers not able to claim at all. 

i completely disagree here. the fact that the campaign is partially funded is suspicious in the first place. depending on what percentage is deposited, it can always happen before the end time. so i wouldn’t use this argument against implementing this method.

> = endTime + 7 days

this works as well, as long as there is this function.

> An alternate solution is to not allow expiration to be 0 in this contract. Wdyt?

don’t like it. it would mean the possibility of having different base configurations between merkle campaigns. and it would restrict the user functionalities

> The test was not needed because until now, clawback and claim cannot happen at the same time. And we have a test for claim failing after the campaign has expired, and another test where clawback fails if the campaign has not expired

i am not sure i understand.
i am talking about  the invariant: `if expiration == 0 → clawback can never be called`

src/SablierMerkleVCA.sol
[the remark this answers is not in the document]
andreivladbrg: >  I personally prefer to use bespoke names for variables and functions than using generic keywords such as timestamps.

start and end timestamps are pretty clear IMO

src/SablierMerkleVCA.sol
smol-ninja: > it would mean the possibility of having different base configurations between merkle campaigns. and it would restrict the user functionalities

No, that wont be required. We can revert in the constructor of MerkleVCA if expiration is 0.

I still think not allowing expiration to be 0 in merkle VCA is a good design choice here. Introducing a separate `transferForgoneAmount` can lead to problems, as I mentioned in a partially funded campaigns. 

I originally posted this question in [issue here](https://github.com/sablier-labs/airdrops/issues/43#issuecomment-2650741073). Would you mind sharing your comment there as well? It seems to be a product design choice to me.

andreivladbrg: > No, that wont be required. We can revert in the constructor of MerkleVCA if expiration is 0.


what i meant is that in `MerkleInstant/MerkleLL/MerkleLT` you can set it to 0 but in `MerkleVCA` you can't. lack of consistency. 

> transferForgoneAmount can lead to problems

it can lead to problems **only** when there is a problematic campaign in the first place (partially funded campaigns). 

> I originally posted this question in https://github.com/sablier-labs/airdrops/issues/43#issuecomment-2650741073. Would you mind sharing your comment there as well? It seems to be a product design choice to me.

that thread looks more about the name of the amount, and not about this discussion we are having here (wether we implement a transfer function). you said "sender can only return it after the expiry" - which assumes there is an expiry, i am talking about when there is not.

smol-ninja: > lack of consistency

So what? MerkleVCA has its own requirement. I don't think being consistent is important here.

> that thread looks more about the name of the amount, and not about this discussion we are having here

[Its relevant because I suggested it there too:](https://github.com/sablier-labs/airdrops/issues/43#issuecomment-2648402297) "However, sender can only return it after the expiry. I first thought of creating a separate function to return sacrificedAmount at any time. But then I realized what if sender does not fully fund the campaign. He can refund the sacrificedAmount early causing loyal users not able to claim at the end of the vesting period." The only part I did not mention is what happens when expiry is 0 to which I am suggesting lets not allow expiry to be 0 in VCA.

src/SablierMerkleVCA.sol
andreivladbrg: you are correct. there are indeed no precision gains. let’s keep the current version of the calculation

smol-ninja: Thank you so much for understanding.

src/SablierMerkleVCA.sol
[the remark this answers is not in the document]
smol-ninja: https://github.com/sablier-labs/airdrops/pull/58/commits/709475bccff0a11bdbea003f6f9680a5f6c10660

src/SablierMerkleVCA.sol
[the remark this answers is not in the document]
smol-ninja: https://github.com/sablier-labs/airdrops/pull/58/commits/709475bccff0a11bdbea003f6f9680a5f6c10660

src/SablierMerkleVCA.sol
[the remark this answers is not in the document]
smol-ninja: I have also replaced vesting with unlock keyword in the context of Merkle VCA to be more explicit. I have replace the variable with `_timestamp`.

https://github.com/sablier-labs/airdrops/pull/58/commits/709475bccff0a11bdbea003f6f9680a5f6c10660

src/types/DataTypes.sol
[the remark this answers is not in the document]
smol-ninja: https://github.com/sablier-labs/airdrops/pull/58/commits/709475bccff0a11bdbea003f6f9680a5f6c10660

src/libraries/Errors.sol
smol-ninja: Changed it to `SablierMerkleVCA_ClaimNotStarted(uint40 startTime)`. I think Period is not required. Its very obvious when it says "claim not started".

https://github.com/sablier-labs/airdrops/pull/58/commits/709475bccff0a11bdbea003f6f9680a5f6c10660

src/SablierMerkleVCA.sol
PaulRBerg: haven't caught up with everything, but I wanted to point out that (as @smol-ninja implied), consistency for consistency's sake is no good.

Default to consistency is OK only when there's no better explanation — and in this case, there is.

andreivladbrg: agree with this argument _against_ consistency

the biggest problem i see with this "partially funded campaign" is the fact that it can happen no matter we implement a `transferForgoneAmount` - so this problem exists in the first place

andreivladbrg: scenario:
- campaign funded with 25%
- 50% of the scheduled time has passed
- 50% of people claim at this exact time
- the campaign runs out of funds, blocking further claims before the forgone amount becomes available for transfer

andreivladbrg: >Whereas, otherwise, sender can claim what "might" belong to the recipient, in case of partially funded campaign.

agree, but that goes against the definition of the forgone amount IMO.
"a sum of tokens that the recipient has voluntarily given up on, allowing the campaign creators to use it as they wish"

you are basically suggesting redistributing it because the campaign creator didn’t fully fund it in the first place

smol-ninja: > you are basically suggesting redistributing it because the campaign creator didn’t fully fund it in the first place

Yes, indirectly.

andreivladbrg: > Yes, indirectly

so, it doesn’t mean that it’s against the definition of the forgone amount—one recipient could say, "i sacrificed my amount with the idea that it goes back to the foundation that created it, and not to the other claimee"

ik, we are both talking about very rare case scenarios

smol-ninja: > "i sacrificed my amount with the idea that it goes back to the foundation that created it, and not to the other claimee

Haha. Very positive way of looking at it. I don't think anyone sacrifices their money because they want it to go to the foundation. I want to prioritise recipients over senders where senders have no way to clawback any amount until the campaign has been expired.

smol-ninja: Included the change in https://github.com/sablier-labs/airdrops/pull/58/commits/156b3b1213a8dbbb65e873c277a9ee8607f3a6bd. And it was a good debate on this topic.

tests/fork/merkle-campaign/MerkleVCA.t.sol
```diff
@@ -0,0 +1,253 @@
...
+        params.timestamps.start = boundUint40(params.timestamps.start, 1, getBlockTimestamp() - 1);
+        params.timestamps.end =
+            boundUint40(params.timestamps.end, params.timestamps.start, MAX_UNIX_TIMESTAMP - 2 weeks);
+
+        // The expiration must exceed the unlock end time by at least 1 week.
+        if (params.timestamps.end > getBlockTimestamp() - 1 weeks) {
+            params.expiration = boundUint40(params.expiration, params.timestamps.end + 1 weeks, MAX_UNIX_TIMESTAMP);
+        } else {
+            // If unlock end time is in the past, set expiration into the future to allow claiming.
+            params.expiration = boundUint40(params.expiration, getBlockTimestamp(), MAX_UNIX_TIMESTAMP);
```
andreivladbrg, line 72: instead of this if block, why not simply do this:

```solidity
params.expiration = boundUint40(params.expiration, params.timestamps.end + 1 weeks, MAX_UNIX_TIMESTAMP);
```

tests/fork/merkle-campaign/MerkleVCA.t.sol
```diff
@@ -0,0 +1,253 @@
...
+
+        vars.merkleVCA.claim{ value: defaults.MINIMUM_FEE() }({
+            index: vars.indexes[params.posBeforeSort],
+            recipient: vars.recipients[params.posBeforeSort],
+            amount: vars.amounts[params.posBeforeSort],
+            merkleProof: vars.merkleProof
+        });
+
+        assertTrue(vars.merkleVCA.hasClaimed(vars.indexes[params.posBeforeSort]));
+
```
andreivladbrg, line 218: why not add `forgoneAmount` assertion?

andreivladbrg (commented):
great work, left some comments re tests:

tests/fork/merkle-campaign/MerkleVCA.t.sol
smol-ninja: Because if `params.timestamps.end` is in the past then we need `params.expiration` to be in the future to test claim function in the fork test.

tests/fork/assets/USDC.t.sol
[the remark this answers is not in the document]
smol-ninja: Good catch.

src/SablierMerkleVCA.sol
```diff
@@ -0,0 +1,144 @@
...
+            params.campaignName,
+            params.expiration,
+            params.initialAdmin,
+            params.ipfsCID,
+            params.merkleRoot,
+            params.token
+        )
+    {
+        // Check: neither unlock start time nor unlock end time is zero.
+        if (params.timestamps.end == 0 || params.timestamps.start == 0) {
```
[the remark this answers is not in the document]
smol-ninja: https://github.com/sablier-labs/airdrops/pull/58/commits/651f9820c600f6766d3d9ff2f54d277df4b964bf

tests/fork/merkle-campaign/MerkleVCA.t.sol
smol-ninja: https://github.com/sablier-labs/airdrops/pull/58/commits/651f9820c600f6766d3d9ff2f54d277df4b964bf

tests/integration/concrete/vca/claim/claim.t.sol
```diff
@@ -0,0 +1,82 @@
...
+
+        merkleVCA.claim{ value: defaults.MINIMUM_FEE() }({
+            index: 1,
+            recipient: users.recipient1,
+            amount: defaults.CLAIM_AMOUNT(),
+            merkleProof: defaults.index1Proof()
+        });
+
+        // It should not update forgone amount.
+        assertEq(merkleVCA.forgoneAmount(), 0, "forgone amount");
```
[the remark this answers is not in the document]
smol-ninja: https://github.com/sablier-labs/airdrops/pull/58/commits/651f9820c600f6766d3d9ff2f54d277df4b964bf

smol-ninja:
Added a new commit, https://github.com/sablier-labs/airdrops/pull/58/commits/651f9820c600f6766d3d9ff2f54d277df4b964bf, incorporating your feedback as well as fixing a bug in VCA fork tests.

tests/integration/concrete/factory/create-merkle-vca/createMerkleVCA.tree
```diff
@@ -0,0 +1,31 @@
...
+   │  └── it should revert
+   └── when start time not zero
+      ├── when end time zero
+      │  └── it should revert
+      └── when end time not zero
+         ├── when end time less than start time
+         │  └── it should revert
+         ├── when end time equals start time
+         │  └── it should revert
+         └── when end time greater than start time
```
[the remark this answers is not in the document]
andreivladbrg: in my mind i had only the case when they equal, but we can also do this:

```solidity
modifier whenEndTimeNotGreater() {
    params = merkleVCAConstructorParams();
    params.timestamps.end = defaults.RANGED_STREAM_START_TIME() - 1;
    _;
    params.timestamps.end = defaults.RANGED_STREAM_START_TIME();
    _;
}

function test_RevertWhen_EndTimeNotGreaterThanStartTime() public whenEndTimeNotGreater {
    // It should revert.
    vm.expectRevert(
        abi.encodeWithSelector(
            Errors.SablierMerkleVCA_StartTimeExceedsEndTime.selector, params.timestamps.start, params.timestamps.end
        )
    );
    merkleFactory.createMerkleVCA({
        params: params,
        aggregateAmount: aggregateAmount,
        recipientCount: recipientCount
    });
}
```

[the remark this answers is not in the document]
smol-ninja: Instead of using a modifier to run tests twice, doesn't it offer better readability with two different branches? Covering all cases through branches leaves very little room to miss out a case even if it duplicates a few lines.

Let me know if it changes your mind about it. Otherwise, I will include your suggestion as technically both are same.

smol-ninja:
@andreivladbrg looking forward to your approval on this PR.

andreivladbrg (approved):
r_downey_thumbs_up

Issue #63, opened 2025-02-10
Move variables from `Defaults` to `Constants`


smol-ninja:
What if we move `constant` and `immutable` variables from `Defaults` to `Constants` contract? This can remove several redundant lines from the code where we have to set the default value onto a local variable to make `expectRevert` work.

For example,

```solidity
uint256 index1 = defaults.INDEX1();
uint128 amount = defaults.CLAIM_AMOUNT();
bytes32[] memory merkleProof = defaults.index1Proof();
uint256 fee = defaults.FEE();

vm.expectRevert(abi.encodeWithSelector(Errors.SablierMerkleBase_InsufficientFeePayment.selector, 0, fee));
merkleBase.claim{ value: 0 }(index1, users.recipient1, amount, merkleProof);
```

would become

```solidity
bytes32[] memory merkleProof = defaults.index1Proof();

vm.expectRevert(abi.encodeWithSelector(Errors.SablierMerkleBase_InsufficientFeePayment.selector, 0, FEE));
merkleBase.claim{ value: 0 }(INDEX1, users.recipient1, CLAIM_AMOUNT, merkleProof);
```

cc @sablier-labs/evm.

andreivladbrg:
i don't remember exactly the initial reason for not inheriting the `Defaults` contract and instead deploying it in `BaseTest`. but now, i believe it would be better to simply inherit it. (cc @PaulRBerg do you remember?)

so, i’d say not to move just some variables to `Constants` while keeping others in `Defaults`, but rather to move all variables to a single contract (either we call it `Constants` or `Defaults`) that will be inherited in `BaseTest` (similarly how we do it in `Flow` tests). wdyt?

smol-ninja:
Yes I agree. I just realized we did the [same in Flow](https://github.com/sablier-labs/flow/blob/main/tests/utils/Constants.sol).

PaulRBerg:
Your plan sounds good.

The original rationale, AFAIK, was to provide context when reading the texts — seeing `defaults.VAR_NAME` instead of `VAR_NAME` points to the fact that that value is a global default used in multiple tests. However, you make a good point @smol-ninja that because of the expect revert behavior of Foundry, we need to pre-define the vars, which makes the original rationale not so strong anymore.

PR #71, opened 2025-02-15
test(refactor): move all variables into Constants contract

smol-ninja:
Closes https://github.com/sablier-labs/airdrops/issues/63


andreivladbrg (commented):
my suggestion in that issue was to completely remove `Defaults` or import it into `Base_Test`, so we don’t have to call anything like `defaults.<foo>`

i don’t see the point of leaving some things in `Defaults` while moving others into `Base_Test`.

so i would rather:
1. move the `index<num>Proof` functions back to `Defaults` and import it into `Base_Test`, or
2. completely move the function from `Defaults` into `Base_Test`.

both work for me

smol-ninja:
Sounds good. I will move everything into Base. That looks a better design to me.

smol-ninja:
Included your suggestion of removing `Defaults` contract @andreivladbrg.

andreivladbrg (commented):
after seeing the actual functions from `Base_Test` now, and at the same time keeping in mind the fact that we will import and use `evm-utils`, and the file `tests/utils/Utils.sol` will be removed after that change

another idea came to mind: what if we move [all these](https://github.com/sablier-labs/airdrops/blob/b9a2fce6d7c3c6017cd816e7666efd5b30d46093/tests/Base.t.sol#L222-L534) functions into `Utils`? IMO they are more of utility functions, especially the compute address functions and `tranchesMerkleLT`.

even though the current version works, just asking as i believe we can improve it (and sorry for so many changes :)) )

smol-ninja:
I didn't what you meant in your first sentence. We will have to keep `Base.t.sol` anyhow because  the `setUp` has some bespoke implementation such as `merkleFactory.setMinimumFee(defaults.MINIMUM_FEE())`.

Also, with respect to moving those function to `Utils`, you pointed to the previous file. After the PR, the Base contains few more functions as you can see [here](https://github.com/sablier-labs/airdrops/blob/tests/default-contract/tests/Base.t.sol).

andreivladbrg:
> I didn't get what you meant in your first sentence. We will have to keep `Base.t.sol` anyhow because the `setUp` has some bespoke implementation such as `merkleFactory.setMinimumFee(defaults.MINIMUM_FEE())`.

i have not said anything about Base.t.sol, but Utils.sol, which [all these functions](https://github.com/sablier-labs/airdrops/blob/b9a2fce6d7c3c6017cd816e7666efd5b30d46093/tests/utils/Utils.sol#L9-L34), can be found in [evm-utils](https://github.com/sablier-labs/evm-utils/blob/main/tests/utils/Utils.sol), thus we will be removing it once we install that here. and now, with this idea in mind (the fact that we - in the future - we will not have a `tests/utils/Utils.sol` anymore) what if we now move [these functions](https://github.com/sablier-labs/airdrops/blob/b9a2fce6d7c3c6017cd816e7666efd5b30d46093/tests/Base.t.sol#L222-L534) (also pointed out in above message) in `tests/utils/Utils.sol`?

> you pointed to the previous file

sorry about that, edited the above message

lmk if i've made self clear now

smol-ninja:
As discussed on Slack, lets leave those functions in Base for now and re-assess it in the `evm-utils` PR.

smol-ninja:
I have resolved the conflicts on this PR and changed the base branch to `refactor/multiple-factories`

tests/Base.t.sol
```diff
@@ -186,33 +226,30 @@ abstract contract Base_Test is Assertions, Constants, DeployOptimized, Modifiers
...
+                                  MERKLE-INSTANT
     //////////////////////////////////////////////////////////////////////////*/
 
     function computeMerkleInstantAddress(
         address campaignCreator,
         address campaignOwner,
         uint40 expiration,
         bytes32 merkleRoot,
-        IERC20 token_
+        IERC20 tokenAddress
```
andreivladbrg, line 242: having type `IERC20` and then saying "Address" isn't ideal IMO, `token_` is better

tests/Base.t.sol
```diff
@@ -237,23 +274,47 @@ abstract contract Base_Test is Assertions, Constants, DeployOptimized, Modifiers
         });
     }
 
+    function merkleInstantConstructorParams(
```
andreivladbrg, line 277: since we have added these in `BaseTest` let's move the mirror functions from `IntegrationTest` here, to group them together

same thing for all `merkle<Campaign>`

tests/Base.t.sol
```diff
@@ -276,23 +337,59 @@ abstract contract Base_Test is Assertions, Constants, DeployOptimized, Modifiers
         });
     }
 
+    function merkleLLConstructorParams(
```
andreivladbrg, line 340: the above comment applies to these functions as well

tests/Base.t.sol
```diff
@@ -276,23 +337,59 @@ abstract contract Base_Test is Assertions, Constants, DeployOptimized, Modifiers
...
+        address campaignOwner,
+        uint40 expiration,
+        ISablierLockup lockupAddress,
```
andreivladbrg, line 343: similar to `tokenAddress` to `token_`

i believe `lockup_` is better

andreivladbrg (commented):
left some comments:

tests/Base.t.sol
smol-ninja: I'd prefer some name without `_`. When writing code with named parameters, its the only inconsistent variable name. Any suggestion is welcome but please not `_` else lets use `_` for other variables too.

tests/Base.t.sol
smol-ninja: https://github.com/sablier-labs/airdrops/pull/71#discussion_r1963703558

Same argument as this. When calling function using named parameters, the DX is a bit worse when one parameter has `_`

smol-ninja: For ex. this looks ugly.

Screenshot 2025-02-20 at 14 49 09

tests/Base.t.sol
andreivladbrg: ok, happy not to declare the params with `_`

smol-ninja: Thanks.

smol-ninja:
gtg. Thanks for the feedback.

Issue #40, opened 2025-01-06
Implement mapping between addresses and claimed streams


PaulRBerg:
Implement the following mapping (name is TBD):

```solidity
mapping(address recipient => uint256[] streamIds) claimed;
```

To make it possible to reverse index streams claimed through an airdrop.

### Discussed in https://github.com/sablier-labs/airdrops/discussions/31

<div type='discussions-op-text'>

<sup>Originally posted by **PaulRBerg** February  1, 2024</sup>

smol-ninja:
I went ahead with `claimedStreamIds` as the variable name because simply calling it `claimed` might confuse with the claimed amount. Feel free to suggest better names otherwise.

PaulRBerg:
How about just `claimedStreams`?

PR #72, opened 2025-02-18
feat: get stream ids associated with airdrop claim

smol-ninja:
Closes https://github.com/sablier-labs/airdrops/issues/40

Depends on https://github.com/sablier-labs/airdrops/pull/71

~I went ahead with `claimedStreamIds` as the variable name because simply calling it `claimed` might confuse with the claimed amount. Feel free to suggest better names otherwise~

Based on https://github.com/sablier-labs/airdrops/issues/40#issuecomment-2668941441, I renamed it to `claimedStream`.

### DIFF
--- a/src/SablierMerkleLL.sol
+++ b/src/SablierMerkleLL.sol
@@ -123,6 +123,9 @@ contract SablierMerkleLL is
             cliffTime
         );
 
+        // Push the stream ID into the `_claimedStreams` array for the recipient.
+        _claimedStreams[recipient].push(streamId);
+
         // Log the claim.
         emit Claim(index, recipient, amount, streamId);
     }
--- a/src/SablierMerkleLL.sol
+++ b/src/SablierMerkleLL.sol
@@ -123,7 +123,7 @@ contract SablierMerkleLL is
             cliffTime
         );
 
-        // Push the stream ID into the `_claimedStreams` array for the recipient.
+        // Effect: push the stream ID into the `_claimedStreams` array for the recipient.
         _claimedStreams[recipient].push(streamId);
 
         // Log the claim.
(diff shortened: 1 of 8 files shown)


src/SablierMerkleLL.sol
```diff
@@ -123,6 +123,9 @@ contract SablierMerkleLL is
             cliffTime
         );
 
+        // Push the stream ID into the `_claimedStreams` array for the recipient.
+        _claimedStreams[recipient].push(streamId);
```
andreivladbrg, line 127: note: CEI will no longer be respected (creating the stream is I, and then pushing is E) buts thats fine

andreivladbrg (approved):
image

src/SablierMerkleLL.sol
smol-ninja: Thats why I didn't use the keyword `Effect` 😆.

BTW any thoughts of not adding `Effect` keyword here? Its obvious that we are updating the state and adding `Effect:` only makes sense when we want to show that we are following CEI, which we are not in this case.

smol-ninja: I am asking that because it becomes `CEIE` haha.

andreivladbrg: > BTW any thoughts on not adding `Effect` keyword here? Its obvious that we are updating the state and adding `Effect:` is only good when we want to show that we are following CEI, which we are not in this case.

so you are suggesting spreading misinformation? 😂

we should take responsibility for not respecting it.

IMO we should keep the keyword

smol-ninja: >  you are suggesting spreading misinformation

Haha. I am suggesting to remove `Check:`, `Effect:`, `Interaction:` keywords when CEI is not respected.

Issue #52, opened 2025-01-30
Update contracts to use the new EVM utils repo


andreivladbrg:
As discussed here https://github.com/sablier-labs/lockup/issues/1125, the repo is created: https://github.com/sablier-labs/evm-utils/

We will need to remove the respective logic from here.

PR #74, opened 2025-02-21
refactor: integrate evm-utils and bump lockup version

smol-ninja:
Closes https://github.com/sablier-labs/airdrops/issues/52

This PR also bumps lockup version.


tests/utils/Constants.sol
```diff
@@ -38,9 +38,6 @@ abstract contract Constants {
 
     // Global
     uint40 internal constant JULY_1_2024 = 1_719_792_000;
```
andreivladbrg: let's update the `JULY_1_2024` to `FEB_1_2025`

tests/utils/DeployOptimized.sol
```diff
@@ -1,15 +1,14 @@
...
-import { StdCheats } from "forge-std/src/StdCheats.sol";
+import { BaseTest as EvmUtilsBase } from "@sablier/evm-utils/src/tests/BaseTest.sol";
 
 import { ISablierMerkleFactoryInstant } from "../../src/interfaces/ISablierMerkleFactoryInstant.sol";
 import { ISablierMerkleFactoryLL } from "../../src/interfaces/ISablierMerkleFactoryLL.sol";
 import { ISablierMerkleFactoryLT } from "../../src/interfaces/ISablierMerkleFactoryLT.sol";
 import { ISablierMerkleFactoryVCA } from "../../src/interfaces/ISablierMerkleFactoryVCA.sol";
 
-abstract contract DeployOptimized is StdCheats, CommonBase {
+abstract contract DeployOptimized is EvmUtilsBase {
```
andreivladbrg: we can only import the `StdCheats` and not the whole `EvmUtilsBase` contract

tests/utils/Modifiers.sol
```diff
@@ -1,11 +1,10 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 pragma solidity >=0.8.22;
 
+import { BaseTest as EvmUtilsBase } from "@sablier/evm-utils/src/tests/BaseTest.sol";
 import { Users } from "./Types.sol";
 
-import { Utils } from "./Utils.sol";
-
-abstract contract Modifiers is Utils {
+abstract contract Modifiers is EvmUtilsBase {
```
andreivladbrg, line 7: This is why I added different contracts for each category (e.g. `Constants`, `Utils`), so they can be imported individually in specific places where only certain functionalities are needed (in this case, only `Utils` would have been required).

But, let’s keep the current version—I don’t have a strong opinion on this

smol-ninja: Yeah I do agree with you.

smol-ninja:
Thanks for the review. The commit looks good.

Issue #37, opened 2025-01-05
Immediately transfer tokens when the vesting period has ended


PaulRBerg:
When `end time <= block timestamp`, transfer the airdrop amount directly to the user, without creating a stream.

cc @sablier-labs/frontend, this may have implications for the informatinal tooltips in the UI

### Discussed in https://github.com/sablier-labs/airdrops/discussions/33

<div type='discussions-op-text'>

<sup>Originally posted by **PaulRBerg** December 23, 2024</sup>

razgraf:
It will have implications to indexers and app logic too. This makes campaigns into a mixture of Instant and Airstreams, since the `Claim` event can now yield either a simple transaction (the transfer, post condition) or a streamId (pre condition).

PaulRBerg:
Right. Would it help if we emitted different events?

razgraf:
Event will be required for sure! But we could simply emit a `Claim` event with `streamId` set to `0` if we don't want to refactor too much. That'll help mark it as a direct transfer instead of a stream claim.

PaulRBerg:
> emit a Claim event with streamId set to 0 if

Yeah, that should work well. I wonder if a boolean flag would be helpful, though? That would be slightly clearer

smol-ninja:
@razgraf since you mentioned setting `streamId` to 0 for direct transfer. To recall, we have the following `Claim` events

1. Merkle Instant: `event Claim(uint256 index, address indexed recipient, uint128 amount)`
2. Merkle Lockups: `event Claim(uint256 index, address indexed recipient, uint128 amount, uint256 indexed streamId)`
3. Merkle VCA: `event Claim(uint256 index, address indexed recipient, uint128 claimableAmount, uint128 totalAmount)`

Do you think it would be better for UI if we just have a common `Claim` event:

`event Claim(uint256 index, address indexed recipient, uint128 amount, uint256 indexed streamId)` with the following conditions:
- `amount` will be `claimedAmount` in case of VCA.
- `streamId` will be `0` for Merkle Instant and Merkle VCA.

razgraf:
It would have simplified things, yes. But since we already support two types of claim events, it doesn't really make a difference right now. So it's a question of simplicity at a contract level.

[...]

PR #77, opened 2025-02-25
feat: transfer tokens directly if claimed after vesting end time

smol-ninja:
Closes https://github.com/sablier-labs/airdrops/issues/37

### Changelog

- Transfers tokens to recipient if claim time exceeds vesting end time in Merkle Lockup
- Adds notes in `claim()`'s natspecs
- Add a new `Claim` event in Merkle Lockup interface as discussed [here](https://github.com/sablier-labs/airdrops/issues/37#issuecomment-2681249131)
- Fuzzes start time in Merkle Lockup's fork tests
- Makes ranged streams as the default streams for integration tests


andreivladbrg (commented):
Just an idea (no strong opinion): now since there is the no guarantee that all `aggregateAmount` is going to be placed `Lockup` contract, should we move the `.approve` function in the `_claim` before creating the stream. The con is obvious, gas increase for the user.

left some comments on `src`:

src/SablierMerkleLL.sol
```diff
@@ -94,38 +94,52 @@ contract SablierMerkleLL is
...
+        // If the stream end time is not in the future, transfer the amount directly to the recipient without creating a
+        // stream.
+        if (timestamps.end <= block.timestamp) {
+            // Interaction: transfer the token.
+            TOKEN.safeTransfer(recipient, amount);
+
+            // Log the claim.
+            emit Claim(index, recipient, amount);
+        }
+        // Otherwise, create the Lockup stream.
```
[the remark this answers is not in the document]
smol-ninja: I intentionally used Lockup and not Lockup Linear. Lockup LL does not make sense anymore as it would be Lockup Lockup Linear.

Since its only contract, imo the stream should be called Lockup stream, created using Linear model. So the short form of that is "create the Lockup stream" and in the next line its clear that `LL` is used due to function signature.

This was my rationale. Wdyt?

src/SablierMerkleLT.sol
```diff
@@ -119,26 +119,38 @@ contract SablierMerkleLT is
...
-            tranches
-        );
-
-        // Effect: push the stream ID into the `_claimedStreams` array for the recipient.
-        _claimedStreams[recipient].push(streamId);
-
-        // Log the claim.
-        emit Claim(index, recipient, amount, streamId);
+        // If the stream end time is not in the future, transfer the amount directly to the recipient without creating a
+        // stream.
```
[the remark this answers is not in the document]
smol-ninja: Good point.

src/SablierMerkleLL.sol
```diff
@@ -94,38 +94,52 @@ contract SablierMerkleLL is
...
+        else {
+            uint40 cliffTime;
+
+            // Calculate cliff time if the cliff duration is greater than 0.
+            if (_schedule.cliffDuration > 0) {
+                cliffTime = timestamps.start + _schedule.cliffDuration;
```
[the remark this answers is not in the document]
smol-ninja: You are right. I will change it. 

I was looking at code of Morpho and they use storage pointers wherever they can to reduce code size. So, I may have been inspired from that while working on the PR.

src/interfaces/ISablierMerkleLockup.sol
```diff
@@ -11,7 +11,10 @@ interface ISablierMerkleLockup is ISablierMerkleBase {
                                        EVENTS
     //////////////////////////////////////////////////////////////////////////*/
 
-    /// @notice Emitted when a recipient claims a Lockup stream.
+    /// @notice Emitted when the recipient receives the airdrop without creating a stream.
```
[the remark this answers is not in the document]
smol-ninja: Your point is valid. But from the point of view of users, being explicit with "without creating a stream" sounds better to me. People always get confused with `claim` and `withdraw` and it might add up more confusion to them.

I am happy to change it though, but if you change you mind lmk.

src/SablierMerkleLL.sol
```diff
@@ -94,38 +94,52 @@ contract SablierMerkleLL is
...
-        );
-
-        // Effect: push the stream ID into the `_claimedStreams` array for the recipient.
-        _claimedStreams[recipient].push(streamId);
-
-        // Log the claim.
-        emit Claim(index, recipient, amount, streamId);
+        // If the stream end time is not in the future, transfer the amount directly to the recipient without creating a
+        // stream.
+        if (timestamps.end <= block.timestamp) {
```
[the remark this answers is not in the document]
smol-ninja: What about `if (schedule.startTime == 0 && timestamps.end <= block.timestamp)`? Can you share the full code you have in your mind?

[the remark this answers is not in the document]
smol-ninja: And BTW its only 2 gas whereas end is 3. So you are removing one `block.timestamp` and adding `&&` should be net plus 1.

smol-ninja:
> Just an idea (no strong opinion): now since there is the no guarantee that all `aggregateAmount` is going to be placed `Lockup` contract, should we move the `.approve` function in the `_claim` before creating the stream. The con is obvious, gas increase for the user.

I understand your rationale and agree with you. But since we have to find a balance between gas efficiency and security, imo this is a very unlikely event where campaign creator passes a malicious lockup and also funds the campaign without checking it. It also means he has not used the Sablier UI to do the same. So thats a low probability security concern.

At the same time, gas usage per claim will [add ~35k ](https://etherscan.io/tx/0x11bc00f3a9e8c63ddabc1e3ffaec5db2235ea4c2b767856e1fc44074b57ea451)which is a lot.

So IMO, the gas increase is not justified here.

src/interfaces/ISablierMerkleLockup.sol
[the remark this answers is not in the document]
andreivladbrg: i have no problem mentioning "without creating a stream" - my point is exactly what you've said: being extra explicit

we are mentioning what it is **NOT** instead of what is actually doing. so lets have both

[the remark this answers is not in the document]
smol-ninja: Ok fair point about prioritising what it is over what it is not

andreivladbrg:
> I understand your rationale and agree with you. But since we have to find a balance between gas efficiency and security, imo this is a very unlikely event where campaign creator passes a malicious lockup and also funds the campaign without checking it. It also means he has not used the Sablier UI to do the same. So thats a low probability security concern.
> 
> At the same time, gas usage per claim will [add ~35k ](https://etherscan.io/tx/0x11bc00f3a9e8c63ddabc1e3ffaec5db2235ea4c2b767856e1fc44074b57ea451)which is a lot.
So IMO, the gas increase is not justified here.

good points, agree, it was just an idea, let's keep the `approve` in the `constructor`

src/SablierMerkleLL.sol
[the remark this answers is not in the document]
andreivladbrg: agree to keep it like this

src/SablierMerkleLL.sol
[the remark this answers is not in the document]
andreivladbrg: ### Benchmark

Ok, i have tested this on remix and i have these results for 
- `foo` = current version
- `bar` = my suggestion

| Condition                              | `foo` Execution Cost (gas) | `bar` Execution Cost (gas) | More Efficient Function |
|----------------------------------------|---------------------------|---------------------------|-------------------------|
| `start = 0` && `end > block.timestamp` | 4841                      | 4989                      | `foo`                   |
| `start > 0` && `end > block.timestamp` | 4931                      | 5092                      | `foo`                   |
| `start > 0` && `end < block.timestamp` | 5102                      | 4941                      | `bar`                   |

#### Results:
- **`foo` is more efficient when** `start = 0` and `end > block.timestamp` (gas cost: 4841 vs. 4989), and when `start > 0` and `end > block.timestamp` (gas cost: 4931 vs. 5092).
- **`bar` is more efficient when** `start > 0` and `end < block.timestamp` (gas cost: 4941 vs. 5102).


<details><summary>Benchmark code I used for testing</summary>
<p>

```solidity
contract A {
    uint256 public startTime;
    uint256 public totalDuration;

    // start = 0  && end > block.timestamp: execution cost 4841 gas
    // start > 0  && end < block.timestamp: execution cost 5102 gas
    function foo() public view returns (uint256) {
        uint256 _startTime;

        if (startTime == 0) {
            _startTime = block.timestamp;
        } else {
            _startTime = startTime;
        }

        uint256 _endTime = _startTime + totalDuration;


        if (_endTime  <= block.timestamp) {
            return 222222222;
        } else {
            return 888888888;
        }

    }

    // start = 0  && end > block.timestamp: execution cost 4989 gas
    // start > 0  && end < block.timestamp: execution cost 4941 gas 
    function bar() public view returns (uint256) {
        uint256 _startTime;

        if (startTime == 0) {
            _startTime = block.timestamp;
        } else {
            _startTime = startTime;
        }

        uint256 _endTime = _startTime + totalDuration;


        if (startTime != 0 && _endTime <= block.timestamp) {
            return 222222222;
        } else {
            return 888888888;
        }

    }

    function setTimes(uint256 _startTime, uint256 _totalDuration) public {
        startTime = _startTime;
        totalDuration = _totalDuration;
    } 
}

``` 

</p>
</details> 


### Should we change the current version?

No, we shouldn't given the results. I no longer think this is a good idea.

src/SablierMerkleLT.sol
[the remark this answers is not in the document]
smol-ninja: https://github.com/sablier-labs/airdrops/pull/77/commits/e6ef7639aa1033f0482b97161eb19ce81bb879c9

src/SablierMerkleLL.sol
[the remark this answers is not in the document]
smol-ninja: https://github.com/sablier-labs/airdrops/pull/77/commits/e6ef7639aa1033f0482b97161eb19ce81bb879c9

Thought the gas saving is only ~450.

src/interfaces/ISablierMerkleLockup.sol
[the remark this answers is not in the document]
smol-ninja: https://github.com/sablier-labs/airdrops/pull/77/commits/e6ef7639aa1033f0482b97161eb19ce81bb879c9

src/SablierMerkleLL.sol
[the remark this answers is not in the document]
smol-ninja: I also ran some tests and would like to share two findings:

1. Same as yours. The current version is more efficient.
2. Second is interesting. Using `block.timestamp` multiple times is more efficient than loading it in memory and then reading the memory from the stack.

| 1st | 2nd |
| -- | -- |
| share with andrei 1 | share with andrei 2 | 

So from this, we should use `block.timestamp` than loading it into memory. I think because its a global variable, its more efficient to read it directly.

smol-ninja:
@andreivladbrg included all your suggestion. Feel free to give it a final review.

src/SablierMerkleLL.sol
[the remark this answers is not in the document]
andreivladbrg: grab the monkey by its tail, lol 😂

> Second is interesting. Using block.timestamp multiple times is more efficient than loading it in memory and then reading the memory from the stack

nice finding, tho, i would add that in our case, usually there is also a casting (either `uint40 --> uint256` or `uint256 --> uint40`) being made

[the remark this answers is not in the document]
smol-ninja: Good point about casting. I suppose in that case using memory variable would be more efficient.

tests/integration/concrete/campaign/ll/claim/claim.tree
```diff
@@ -1,17 +1,21 @@
 Claim_MerkleLL_Integration_Test
 └── when Merkle proof valid
-   ├── when total percentage greater than 100 
+   ├── when total percentage greater than 100
    │  └── it should revert
    └── when total percentage not greater than 100
```
andreivladbrg, line 5: let’s move this branch after `when vesting end time not exceed claim time`

the reason is that even if the schedule is incorrectly set up, if the end time is greater than `block.timestamp`, the claim can still be made

tests/integration/concrete/campaign/lt/claim/claim.tree
```diff
@@ -6,9 +6,13 @@ Claim_MerkleLT_Integration_Test
    │  └── when total percentage greater than 100
    │     └── it should revert
    └── when total percentage 100
```
andreivladbrg, line 8: same as above

tests/integration/concrete/campaign/lt/claim/claim.tree
```diff
@@ -6,9 +6,13 @@ Claim_MerkleLT_Integration_Test
...
       │  └── it should emit a {Claim} event
-      └── when stream start time not zero
-         ├── it should create a stream with `STREAM_START_TIME` as start time
-         └── it should emit a {Claim} event
+      └── when vesting end time exceeds claim time
+         ├── when stream start time zero
+         │  ├── it should create a stream with block.timestamp as start time
+         │  └── it should emit a {Claim} event
+         └── when stream start time not zero
+            ├── it should create a stream with `RANGED_STREAM_START_TIME` as start time
```
andreivladbrg, line 17: i believe using variable names from tests in the tree files is not a good idea.

these files should be understood without reading the test logic. without more context, `RANGED_STREAM_START_TIME` is an unknown

tests/fork/merkle-campaign/MerkleLL.t.sol
```diff
@@ -191,56 +224,68 @@ abstract contract MerkleLL_Fork_Test is Fork_Test {
...
+        else {
+            vars.expectedUnlockAmounts.start =
+                ud60x18(vars.amounts[params.posBeforeSort]).mul(START_PERCENTAGE.intoUD60x18()).intoUint128();
+            vars.expectedUnlockAmounts.cliff =
+                ud60x18(vars.amounts[params.posBeforeSort]).mul(CLIFF_PERCENTAGE.intoUD60x18()).intoUint128();
+
+            // Assert that the stream has been created successfully.
+            assertEq(lockup.getCliffTime(vars.expectedStreamId), vars.expectedStartTime + CLIFF_DURATION, "cliff time");
+            assertEq(
+                lockup.getDepositedAmount(vars.expectedStreamId), vars.amounts[params.posBeforeSort], "deposited amount"
```
andreivladbrg, line 253: these assertions become quite long. what if we use the function declared in [lockup/tests/Assertions](https://github.com/sablier-labs/lockup/blob/723a95675a16a2b94f5c1140af848d09f30421d2/tests/utils/Assertions.sol#L94-L97)?

maybe it would reduce the code

tests/fork/merkle-campaign/MerkleLT.t.sol
```diff
@@ -189,44 +223,58 @@ abstract contract MerkleLT_Fork_Test is Fork_Test {
...
+            );
+        }
+        // Assertions when vesting end time exceeds the block time.
+        else {
+            assertEq(
+                lockup.getDepositedAmount(vars.expectedStreamId), vars.amounts[params.posBeforeSort], "deposited amount"
+            );
+            assertEq(lockup.getEndTime(vars.expectedStreamId), vars.expectedStartTime + TOTAL_DURATION, "end time");
+            assertEq(lockup.getLockupModel(vars.expectedStreamId), Lockup.Model.LOCKUP_TRANCHED);
+            assertEq(lockup.getRecipient(vars.expectedStreamId), vars.recipients[params.posBeforeSort], "recipient");
```
andreivladbrg, line 249: same as before.

andreivladbrg (changes requested):
left some more comments re `tests`

if you agree with them, i can push a commit with the changes, as i implemented them while reviewing

tests/integration/concrete/campaign/ll/claim/claim.tree
smol-ninja: Are you suggesting the following?

```
# when vesting end time not exceed claim time
## when total percentage not greater than 100
```

If so, I didn't understand your rationale.

tests/integration/concrete/campaign/lt/claim/claim.tree
smol-ninja: Agree with you. I could have said `it should create a ranged stream` which implicitly implies that `RANGED_STREAM_START_TIME` is not zero.

tests/integration/concrete/campaign/ll/claim/claim.tree
andreivladbrg: > Are you suggesting the following?

yes

> If so, I didn't understand your rationale.

you can have the `total percentage not greater than 100` (schedule is incorrectly set up) and still claim the airdrop `if vesting end time exceeds block.timestamp`

tests/fork/merkle-campaign/MerkleLL.t.sol
smol-ninja: Feel free to make the change if it reduces the code drastically. But if its just 1-2 lines, I am fine with current version too as I don't have problem with so many assertions as they are important for tests.

tests/integration/concrete/campaign/ll/claim/claim.tree
smol-ninja: Can you given an example of incorrect schedule where total percentage not greater than 100? I am not able to understand your point. IMO total percentage <= 100 is always valid. When you say incorrect schedule, what do you mean?

The reason `when vesting end time exceeds claim time` is below `total percentage not greater than 100` is because airdrop claim always work as long as total percentage <= 100.

andreivladbrg: > The reason `when vesting end time exceeds claim time` is below `total percentage not greater than 100` is because airdrop claim always work as long as total percentage <= 100.

if `endTime > block.timestamp` that is irrelevant

> Can you given an example of incorrect schedule where total percentage not greater than 100? I am not able to understand your point. IMO total percentage <= 100 is always valid. When you say incorrect schedule, what do you mean?

exactly this: https://github.com/sablier-labs/airdrops/blob/e6ef7639aa1033f0482b97161eb19ce81bb879c9/tests/integration/concrete/campaign/ll/claim/claim.t.sol#L22-L37

if you add this line before `claim`, it won't revert as expected in the test above:
```solidity
// Forward in time to the end of the vesting period.
vm.warp({ newTimestamp: RANGED_STREAM_END_TIME + 1 });

```

tests/fork/merkle-campaign/MerkleLT.t.sol
andreivladbrg: 06404a51f23fb2b8d00db59facb3dd6d9e2ec95d

tests/integration/concrete/campaign/lt/claim/claim.tree
andreivladbrg: 06404a51f23fb2b8d00db59facb3dd6d9e2ec95d

tests/integration/concrete/campaign/lt/claim/claim.tree
```diff
@@ -1,18 +1,18 @@
...
+      │  │  └── it should revert
+      │  └── when total percentage greater than 100
+      │     └── it should revert
+      └── when total percentage 100
          ├── when stream start time zero
          │  ├── it should create a stream with block.timestamp as start time
          │  └── it should emit a {Claim} event
          └── when stream start time not zero
-            ├── it should create a stream with `RANGED_STREAM_START_TIME` as start time
+            ├── it should create a stream with the correct start time
```
smol-ninja, line 17: All start time is correct. Why not say that "it should create a ranged stream with provided start time"?

In the above branch we say `it should create a stream with block.timestamp as start time` or it makes sense to say it "it should create a ranged stream with provided start time" here.

tests/integration/concrete/campaign/ll/claim/claim.tree
smol-ninja: Good catch. Its clear now.

tests/integration/concrete/campaign/lt/claim/claim.tree
andreivladbrg: sure

smol-ninja: https://github.com/sablier-labs/airdrops/pull/77/commits/df60eaa832a3de6ddfeb89821fda686d12ffd25a

andreivladbrg (approved):
shogun_lgtm

PR #79, opened 2025-02-27
docs: update license

Radovenchyk:


smol-ninja:
Thank you for the PR @Radovenchyk. The same change can also be made in [Lockup](https://github.com/sablier-labs/lockup/blob/main/LICENSE.md) and [Flow](https://github.com/sablier-labs/flow/blob/main/LICENSE.md) repos.

Would you like to open the same PRs in those two repos as well?

Radovenchyk:
Ok

PR #82, opened 2025-03-04
test: dry-fy fork tests

### LINKED ISSUE #81: DRY-ify fork tests
Similar to the [lockup fork tests](https://github.com/sablier-labs/lockup/blob/staging/tests/fork/Lockup.t.sol), we can dry-fy the Merkle campaign fork tests code as well.  

For example:  
- [MerkleInstant code](https://github.com/sablier-labs/airdrops/blob/b9dba976afb53d3c6d8020236743732909377fc7/tests/fork/merkle-campaign/MerkleInstant.t.sol#L71-L108)  
- [MerkleLL code](https://github.com/sablier-labs/airdrops/blob/b9dba976afb53d3c6d8020236743732909377fc7/tests/fork/merkle-campaign/MerkleLL.t.sol#L91-L128)  

This is just an example—there are more lines that can be DRY-ified.

andreivladbrg:
Closes #81

### DIFF
--- a/tests/Base.t.sol
+++ b/tests/Base.t.sol
@@ -12,6 +12,7 @@ import { SablierLockup } from "@sablier/lockup/src/SablierLockup.sol";
 import { LockupTranched } from "@sablier/lockup/src/types/DataTypes.sol";
 import { Merkle } from "murky/src/Merkle.sol";
 import { ISablierMerkleBase } from "src/interfaces/ISablierMerkleBase.sol";
+import { ISablierMerkleFactoryBase } from "src/interfaces/ISablierMerkleFactoryBase.sol";
 import { ISablierMerkleFactoryInstant } from "src/interfaces/ISablierMerkleFactoryInstant.sol";
 import { ISablierMerkleFactoryLL } from "src/interfaces/ISablierMerkleFactoryLL.sol";
 import { ISablierMerkleFactoryLT } from "src/interfaces/ISablierMerkleFactoryLT.sol";
@@ -52,6 +53,10 @@ abstract contract Base_Test is Assertions, Constants, DeployOptimized, Merkle, M
     //////////////////////////////////////////////////////////////////////////*/
 
     ISablierLockup internal lockup;
+    /// @dev A test contract meant to be overridden by the implementing Merkle campaign contracts.
+    ISablierMerkleBase internal merkleBase;
+    /// @dev A test contract meant to be overridden by the implementing Merkle factory contracts.
+    ISablierMerkleFactoryBase internal merkleFactoryBase;
     ISablierMerkleFactoryInstant internal merkleFactoryInstant;
     ISablierMerkleFactoryLL internal merkleFactoryLL;
     ISablierMerkleFactoryLT internal merkleFactoryLT;
--- a/tests/fork/merkle-campaign/MerkleBase.t.sol
+++ b/tests/fork/merkle-campaign/MerkleBase.t.sol
@@ -1,0 +1,200 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.22 <0.9.0;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { Arrays } from "@openzeppelin/contracts/utils/Arrays.sol";
+
+import { ISablierMerkleBase } from "src/interfaces/ISablierMerkleBase.sol";
+import { ISablierMerkleFactoryBase } from "src/interfaces/ISablierMerkleFactoryBase.sol";
+
+import { MerkleBuilder } from "./../../utils/MerkleBuilder.sol";
+import { Fork_Test } from "./../Fork.t.sol";
+
+abstract contract MerkleBase_Fork_Test is Fork_Test {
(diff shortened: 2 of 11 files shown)


tests/Base.t.sol
```diff
@@ -52,6 +53,10 @@ abstract contract Base_Test is Assertions, Constants, DeployOptimized, Merkle, M
     //////////////////////////////////////////////////////////////////////////*/
 
     ISablierLockup internal lockup;
+    /// @dev A test contract meant to be overridden by the implementing Merkle campaign contracts.
```
smol-ninja, line 56: From Solidity docs:

> Documentation is inserted above each contract, interface, library, function, and event. A public state variable is equivalent to a function for the purposes of NatSpec.

Thus, we should avoid natspec format for internal and private variables, and simply use comments.

Suggestion

```solidity
    // The following two variables are meant to be overridden by the implementing Merkle campaign and Merkle factory
    // contracts, respectively.
    ISablierMerkleBase internal merkleBase;
    ISablierMerkleFactoryBase internal merkleFactoryBase;
```

tests/fork/merkle-campaign/Merkle.t.sol
```diff
@@ -0,0 +1,200 @@
...
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { Arrays } from "@openzeppelin/contracts/utils/Arrays.sol";
+
+import { ISablierMerkleBase } from "src/interfaces/ISablierMerkleBase.sol";
+import { ISablierMerkleFactoryBase } from "src/interfaces/ISablierMerkleFactoryBase.sol";
+
+import { MerkleBuilder } from "./../../utils/MerkleBuilder.sol";
+import { Fork_Test } from "./../Fork.t.sol";
+
+abstract contract Merkle_Fork_Test is Fork_Test {
```
smol-ninja, line 13: I think we can move all of this logic into `Fork_Test`. The purpose of the `Fork_Test` is to act as a parent contract for the fork tests.

I don't see a point in having two shared files.

tests/fork/merkle-campaign/Merkle.t.sol
```diff
@@ -0,0 +1,200 @@
...
+        LeafData[] leafData;
+        uint256 posBeforeSort;
+    }
+
+    struct Vars {
+        uint256 initialAdminBalance;
+        uint256 aggregateAmount;
+        uint128[] amounts;
+        uint128 clawbackAmount;
+        address expectedMerkle;
```
smol-ninja, line 39: merkle is a very generic term and may mean a lot of things. Should we call it `expectedMerkleCampaign`?

smol-ninja (changes requested):
350 less lines feels like a win. Nice work and few comments below:

tests/fork/merkle-campaign/Merkle.t.sol
andreivladbrg: > I think we can move all of this logic into `Fork_Test`

I believe its better in the current version:
- `Fork_Test` is very generic fork set up
- `Merkle_Fork_Test` contains very specific things to the merkle functionalities + actual common fuzz/tests/assertions

also, you went for the [same approach in lockup](https://github.com/sablier-labs/lockup/blob/3d6c92bc8659ded21740f930099bcb11786e14a2/tests/fork/Lockup.t.sol#L13) - which i agree with (we could technically move the `Lockup_Fork_Test` logic in `Fork_Test`

tests/fork/merkle-campaign/Merkle.t.sol
andreivladbrg: sounds good.

tests/Base.t.sol
andreivladbrg: given we are talking about tests, that's kinda irrelevant since we're not publishing/deploying the tests anywhere, IMO

also, if we have natspec comments for internal/private declarations under `src` — IMO, we should keep them. for someone reading the code, when you see a nicely formatted comment (with a different color), it kind of psychologically triggers you to read it :)) — or at least makes it more likely that you don't skip it e.g.:

image

vs

image

also, from solidity docs:

Screenshot 2025-03-10 at 3 37 53 PM

smol-ninja: Agree with your argument about reading experience and color. Lets keep it.

tests/fork/merkle-campaign/Merkle.t.sol
smol-ninja: OK lets keep it this way.

How about renaming `Merkle_Fork_Test` to `MerkleAirdrops_Fork_Test` or `AIrdrops_Fork_Test`. `Merkle_Fork_Test` sounds like a fork test for Merkle tree core logic, since Merkle is a very generic term.

andreivladbrg: `MerkleCampaigns_Fork_Test` sounds good, similar to the comment below 👍

andreivladbrg: as discussed on slack, i renamed it to `MerkleBase` (both the contract and the file)

smol-ninja (approved):
pikachu

Issue #105, opened 2025-03-20
Getter for calculating the claim amount


PaulRBerg:
In the VCA contract, we absolutely need a new getter for calculating the claim amount. This logic needs to be moved to a subroutine:

https://github.com/sablier-labs/airdrops/blob/41f2e742744caabd64c388a32aa6eaf26bbe4654/src/SablierMerkleVCA.sol#L125-L130

smol-ninja:
Can you provide the interface for the getter function that you have in your mind? Would it contain merkle proof for amount verification?

PaulRBerg:
I've updated the OP to include the interface.

Why would it contain the Merkle proof for verification? This is just an informational getter.

smol-ninja:
> Why would it contain the Merkle proof for verification? This is just an informational getter

Why would a user use this getter? To find how much he would receive if he claims right now? However, the amount is not stored onchain. So, its not possible for any user to know how much he would receive if he claims right now using this function. Therefore, I do not see why would something use this with a random amount.

Now for the UI, which has access to amounts, can use this function but they can also use `timestamps()` getter and run a simple calculation to show how much user would receive they he claims right now.

Who do you think would this function be really helpful to?

PaulRBerg:
The Sablier UI might find this getter helpful, yes. At a minimum, it would serve as an excellent safety check for the calculations performed offchain. @sablier-labs/frontend, could you please confirm?

That said, a much more important audience are the users themselves. What happens if the Sablier UI goes down? The user might remember their full amount and want to know how much they can claim now!

Or, if someone wishes to build a third-party airdrop checker, they would find this getter helpful (note: this checker would be good for Sablier, since they would send users our way!)

Without this getter, there would be no way to easily calculate the claim amount, and this would be obviously bad for an airdrop product.

> the amount is not stored onchain

The full amount is not stored onchain, but it's typically communicated through all sorts of channels — claim pages, Google Sheets spreadsheets, airdrop checkers, etc.

PaulRBerg:
In short; implementing this getter takes us little development effort, but opens up several potential avenues for BD.

smol-ninja:
The timestamps that we were discussing about returns vesting start and end time. I think you meant a timestamp in future to simulate claimable amount.

andreivladbrg:
Agree with the new getter.  
In the meantime, if there's an urgent need for an on-chain getter, the `StreamingMath` library can be used, with the correct params are passed.

[...]

Issue #108, opened 2025-03-20
Getter for calculating forgone amount


PaulRBerg:
Similar to https://github.com/sablier-labs/airdrops/issues/105 but for the foregone amount:

```solidity
function calculateForgoneAmount(uint128 fullAmount, uint40 claimTime) external view returns (uint128);
```

smol-ninja:
When `claimTime <= startTime`, what should `calculateForgoneAmount` return? Zero or should it revert?

smol-ninja:
When `claimTime <= startTime`, what should `calculateForgoneAmount` return? Full amount or should it revert?

IMO it should revert.

PaulRBerg:
It should return zero in that case.

Nothing would be forgone because the claim function wouldn't allow the user to claim anything. It's a no-op.

smol-ninja:
Makes sense.

PR #112, opened 2025-03-21
refactor: PRB feedback

PaulRBerg:
## RFC

The PR is not yet complete but you can start reviewing @smol-ninja @andreivladbrg.

I'd appreciate your help with implementing https://github.com/sablier-labs/airdrops/issues/108, https://github.com/sablier-labs/airdrops/issues/105, and https://github.com/sablier-labs/airdrops/issues/111.

## Issues Closed

- Closes https://github.com/sablier-labs/airdrops/issues/122
- Closes https://github.com/sablier-labs/airdrops/issues/97
- Closes https://github.com/sablier-labs/airdrops/issues/98
- Closes https://github.com/sablier-labs/airdrops/issues/99
- Closes https://github.com/sablier-labs/airdrops/issues/100
- Closes https://github.com/sablier-labs/airdrops/issues/101
- Closes https://github.com/sablier-labs/airdrops/issues/102
- Closes https://github.com/sablier-labs/airdrops/issues/103

PR #115, opened 2025-03-22
Getter functions in VCA

andreivladbrg:
Closes https://github.com/sablier-labs/airdrops/issues/105 and https://github.com/sablier-labs/airdrops/issues/108

**Notes:** 
- I implemented the internal `_calculateClaimAmount` so that it is optimized for `_claim`. 
- I could have also marked `calculateClaimAmount` as `external` and added additional logic in `calculateForgoneAmount` to make it more gas-efficient. But honestly, I didn't find it worth it given the minimal gas savings.

### Tasks

- [x] Return 0 in `calculateForgoneAmount` when `claimTime <= startTime`


smol-ninja:
@andreivladbrg we need to create these PRs against https://github.com/sablier-labs/airdrops/pull/112. There are [`TODO`](https://github.com/sablier-labs/airdrops/blob/090a1256c2f82026b7d4b914ca3f5ebe91bc5bc1/src/SablierMerkleVCA.sol#L102) in the code already.

I will also update https://github.com/sablier-labs/airdrops/pull/116.

andreivladbrg:
@smol-ninja @PaulRBerg updated the branch

smol-ninja:
@andreivladbrg do you want to bump evm-utils and rename `resetPrank` in this PR? So that we can merge it into https://github.com/sablier-labs/airdrops/pull/112 after review. I will make changes to my PR once this is merged into https://github.com/sablier-labs/airdrops/pull/112.

andreivladbrg:
> @andreivladbrg do you want to bump evm-utils and rename `resetPrank` in this PR? So that we can merge it into #112 after review. I will make changes to my PR once this is merged into #112.

sure, will do it now

tests/integration/concrete/campaign/vca/calculate-claim-amount/calculateClaimAmount.tree
```diff
@@ -0,0 +1,7 @@
+CalculateClaimAmount_MerkleVCA_Integration_Test
+├── when start time in the future
+│  └── it should return 0
+├── when end time in past
```
smol-ninja, line 4: it should be

```
# when start time not in the past
## it should return 0
# when start time in the past
## when end time not in future
### it should return the full amount
## when end time in future
### it should calculate the claim amount
```

tests/integration/concrete/campaign/vca/calculate-forgone-amount/calculateForgoneAmount.tree
smol-ninja, line 1: Similar changes here.

smol-ninja (changes requested):

src/SablierMerkleVCA.sol
```diff
@@ -98,17 +98,18 @@ contract SablierMerkleVCA is
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @inheritdoc ISablierMerkleVCA
-    function calculateClaimAmount(uint128 fullAmount) external pure returns (uint128) {
-        // TODO: implement
-        fullAmount;
-        return 0;
+    function calculateClaimAmount(uint128 fullAmount) public view returns (uint128) {
```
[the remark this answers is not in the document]
andreivladbrg: see OP

src/SablierMerkleVCA.sol
```diff
@@ -98,17 +98,18 @@ contract SablierMerkleVCA is
...
+        // If the schedule start time is in the future, the claimable amount is zero.
+        if (_schedule.startTime >= block.timestamp) {
```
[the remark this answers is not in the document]
andreivladbrg: see note in OP

`_calculateClaimAmount`  its meant to be optimized for `_claim`, so that it won't make the check twice

src/SablierMerkleVCA.sol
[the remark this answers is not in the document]
smol-ninja: Sorry I missed that. Just realised that the function should be changed as per https://github.com/sablier-labs/airdrops/issues/105#issuecomment-2748266644.

src/SablierMerkleVCA.sol
```diff
@@ -117,38 +123,49 @@ contract SablierMerkleVCA is
...
+
+        // Calculate the claim.
+        uint128 claimAmount = _calculateClaimAmount(fullAmount, blockTimestamp);
+
+        uint128 forgoneAmount;
 
-            // Effect: update the forgone amount.
+        // Effect: update the total forgone amount.
+        if (claimAmount < fullAmount) {
             forgoneAmount = fullAmount - claimAmount;
```
smol-ninja: Can be moved under `unchecked`

src/SablierMerkleVCA.sol
```diff
@@ -117,38 +123,49 @@ contract SablierMerkleVCA is
...
-        if (vestingSchedule.endTime <= blockTimestamp) {
+        // Calculate the claim amount.
+        if (vestingSchedule.endTime <= claimTime) {
             // If the vesting period has ended, the recipient can claim the full amount.
-            claimAmount = fullAmount;
+            return fullAmount;
         } else {
             // Otherwise, calculate the claimable amount based on the elapsed time.
-            uint40 elapsedTime = blockTimestamp - vestingSchedule.startTime;
+            uint40 elapsedTime = claimTime - vestingSchedule.startTime;
```
smol-ninja: This can be moved under `unchecked`

smol-ninja (approved):
Please find my feedback below which I have addressed in https://github.com/sablier-labs/airdrops/pull/115/commits/e94dfa87727e0b4e6ae037e0ea7556bc872c5c19.

LMK if you agree.

PaulRBerg:
I see there are now conflicts .. apologies @andreivladbrg. They are due to the `FactoryMerkle` refactor.

I will wait for you to update this PR and the other before (i) reviewing my PR and (ii) reviewing your PRs and (iii) making any additional changes.

src/SablierMerkleVCA.sol
```diff
@@ -98,17 +98,23 @@ contract SablierMerkleVCA is
...
-        fullAmount;
-        return 0;
+    function calculateClaimAmount(uint128 fullAmount, uint40 claimTime) public view returns (uint128) {
+        // If zero is passed as the claim time, use the block time.
+        if (claimTime == 0) {
+            claimTime = uint40(block.timestamp);
+        }
+
+        // If the schedule start time is greater than claim time, the claimable amount is zero.
+        if (_schedule.startTime >= claimTime) {
```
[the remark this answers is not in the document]
andreivladbrg: As said, it is optimized for `_claim` — you can't claim with `_schedule.startTime >= claimTime`; it would revert.

Also, we have these duplicated now.  

```solidity
// Zero is used a sentinel value for `block.timestamp`.
if (claimTime == 0) {
    claimTime = uint40(block.timestamp);
}
```

There is no benefit to making the function `external` instead of `public`, as this function is read-only. Since it's gasless, the difference of 10 gas or so doesn't really matter.

It's better to have a single point of failure (a common logic function).

> When claimTime <= _schedule.startTime, calculateForgoneAmount would revert with overflow.

I think that wouldn't be as clear as a return value.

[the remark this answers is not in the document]
smol-ninja: > As said, it is optimized for _claim — you can't claim with _schedule.startTime >= claimTime; it would revert.

Can you please explain this? You cannot claim when `_schedule.startTime >= claimTime` as it reverts. Thats the current behaviour.

> we have these duplicated now

Agree and made calculateClaimAmount public again.

And whats the issue with the changes where I have moved `claimTime <= _schedule.startTime` logic into `_calculateClaimAmount`?

smol-ninja:
Pasting the following from a discussion between me and Andrei (I think it requires some thoughts):

I found an issue that there is a possibility that `claimAmount` is 0 even when `startTime < claimTime`, especially when allocation is such that it takes `> 1` second to stream 1 wei of token. In that case, user can lose 100% of the amount if claimed after claim time but within that window. Therefore, I think it makes more sense to compare `claimAmount == 0` and revert than comparing `startTime >= claimTime` and revert.

We can also introduce two errors: `claimAmountZero` alongside `claimNotStarted`. Thoughts @sablier-labs/evm?

src/SablierMerkleVCA.sol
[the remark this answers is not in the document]
smol-ninja: I have reverted some changes in https://github.com/sablier-labs/airdrops/pull/115/commits/604427bbb64a3f7ffc6fdd46352396426078bf08 as per https://github.com/sablier-labs/airdrops/issues/124 which will solve zero claim amount problem but had to use `_calculateClaimAmount` in `calculateForgoneAmount` as per https://github.com/sablier-labs/airdrops/issues/108#issuecomment-2749038687.

Let me know if there a better solution.

smol-ninja:
The [chainlink fork tests](https://github.com/sablier-labs/airdrops/blob/staging/tests/fork/chainlink-oracle/ChainlinkOracle.t.sol) have started to fail frequently: [see latest run](https://github.com/sablier-labs/airdrops/actions/runs/14045484412/job/39325339239?pr=115).

PaulRBerg:
The failed tests on BSC might be because of an RPC issue .. we encountered a similar issue in the frontend today. Maybe let's use Infura for BSC?

smol-ninja:
Its already using infura for BSC: [foundry.toml](https://github.com/sablier-labs/airdrops/blob/staging/foundry.toml#L96)

src/SablierMerkleVCA.sol
```diff
@@ -117,40 +134,56 @@ contract SablierMerkleVCA is
...
+
+        // Calculate the claim.
+        uint128 claimAmount = _calculateClaimAmount(fullAmount, blockTimestamp);
+        uint128 forgoneAmount;
 
-            // Effect: update the forgone amount.
-            forgoneAmount = fullAmount - claimAmount;
-            totalForgoneAmount += forgoneAmount;
+        // Effect: update the total forgone amount.
+        if (claimAmount < fullAmount) {
```
PaulRBerg, line 182: Let's add an assertion that `claimAmount == fullAmount`

```solidity
assert(claimAmount == fullAmount);
```

tests/utils/Constants.sol
```diff
@@ -15,6 +15,7 @@ abstract contract Constants {
     uint256 public constant MIN_FEE_WEI = (1e18 * MIN_FEE_USD) / 3000e8; // at $3000 per ETH price
     uint128 public constant START_AMOUNT = 100e18;
     UD2x18 public immutable START_PERCENTAGE = (ud(START_AMOUNT).div(ud(CLAIM_AMOUNT)).intoUD2x18());
+    uint128 internal constant VCA_FULL_AMOUNT = CLAIM_AMOUNT;
```
PaulRBerg, line 18: Has to be removed

src/SablierMerkleVCA.sol
```diff
@@ -117,40 +134,56 @@ contract SablierMerkleVCA is
...
-            uint40 totalDuration = vestingSchedule.endTime - vestingSchedule.startTime;
+            unchecked {
+                elapsedTime = claimTime - vestingSchedule.startTime;
+                totalDuration = vestingSchedule.endTime - vestingSchedule.startTime;
+            }
 
             // Safe to cast because the division results into a value less than `fullAmount`, which is already an
             // `uint128`.
-            claimAmount = uint128((uint256(fullAmount) * elapsedTime) / totalDuration);
+            return uint128((uint256(fullAmount) * elapsedTime) / totalDuration);
```
PaulRBerg, line 160: Why not do this instead?

```suggestion
            return fullAmount * uint128(elapsedTime) / uint128(totalDuration);
```

tests/utils/Constants.sol
smol-ninja: I added it because I am using it in `calculateClaimAmount.t.sol` and `calculateForgoneAmount.t.sol`. Please have a look and let me know if it makes sense. I could have used the `CLAIM_AMOUNT` there but `VCA_FULL_AMOUNT` makes more sense since we are already using it in `claim`.

src/SablierMerkleVCA.sol
smol-ninja: Because now both `elapsedTime` and `fullAmount` are `uint128`, the result would be of type `uint128`. However, during multiplication, the result can exceed `uint128` until the division is performed. Therefore, `fullAmount` was casted to `uint256` to accommodate multiplication result in case it exceeds `uint128`.  

So, its not dangerous cast.

PaulRBerg: You're right!

tests/utils/Constants.sol
PaulRBerg: Oh, right

src/SablierMerkleVCA.sol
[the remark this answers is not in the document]
andreivladbrg: we can keep like this 👍

Issue #111, opened 2025-03-21
Turn "timestamps" into "START_TIME" and "END_TIME"


PaulRBerg:
As discussed [here](https://github.com/sablier-labs/airdrops/issues/106#issuecomment-2741782757), it'd be cleaner for the `schedule` struct in `MerkleVCA` to be refactored into separate `START_TIME` and `END_TIME` values.

PR #116, opened 2025-03-22
refactor: MerkleVCA timestamps into state variables

smol-ninja:
Closes https://github.com/sablier-labs/airdrops/issues/111

### DIFF
--- a/src/types/DataTypes.sol
+++ b/src/types/DataTypes.sol
@@ -126,27 +126,21 @@ library MerkleLT {
 library MerkleVCA {
     /// @notice Struct encapsulating the constructor parameters of Merkle VCA campaigns.
     /// @param campaignName The name of the campaign.
+    /// @param endTime Vesting end time.
     /// @param expiration The expiration of the campaign, as a Unix timestamp.
     /// @param initialAdmin The initial admin of the campaign.
     /// @param ipfsCID The content identifier for indexing the contract on IPFS.
     /// @param merkleRoot The Merkle root of the claim data.
-    /// @param schedule Struct encapsulating the vesting start time and end time.
+    /// @param startTime Vesting start time.
     /// @param token The contract address of the ERC-20 token to be distributed.
     struct ConstructorParams {
         string campaignName;
+        uint40 endTime;
         uint40 expiration;
         address initialAdmin;
         string ipfsCID;
         bytes32 merkleRoot;
-        Schedule schedule;
-        IERC20 token;
-    }
-
-    /// @notice Struct encapsulating the vesting start time and end time.
-    /// @param startTime The timestamp when the airdrop token begins to unlock.
-    /// @param endTime The timestamp when the airdrop token unlocks 100%.
-    struct Schedule {
         uint40 startTime;
-        uint40 endTime;
+        IERC20 token;
     }
 }
--- a/src/types/DataTypes.sol
+++ b/src/types/DataTypes.sol
@@ -17,6 +17,7 @@ library FactoryMerkle {
 
 library MerkleInstant {
     /// @notice Struct encapsulating the constructor parameters of Merkle Instant campaigns.
+    /// @dev The fields are arranged alphabetically.
     /// @param campaignName The name of the campaign.
     /// @param expiration The expiration of the campaign, as a Unix timestamp. A value of zero means the campaign does
     /// not expire.
@@ -36,6 +37,7 @@ library MerkleInstant {
 
 library MerkleLL {
     /// @notice Struct encapsulating the constructor parameters of Merkle Lockup Linear campaigns.
+    /// @dev The fields are arranged alphabetically.
     /// @param campaignName The name of the campaign.
     /// @param cancelable Indicates if the Lockup stream will be cancelable after claiming.
     /// @param expiration The expiration of the campaign, as a Unix timestamp. A value of zero means the campaign does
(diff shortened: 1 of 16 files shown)


PaulRBerg:
Thanks Shub, unfortunately there will likely be conflicts with https://github.com/sablier-labs/airdrops/pull/112. Could you target that PR's branch instead?

smol-ninja:
@PaulRBerg it would be better to merge this into https://github.com/sablier-labs/airdrops/pull/115 and then review https://github.com/sablier-labs/airdrops/pull/115 as a whole.

PaulRBerg:
Sorry @smol-ninja I've missed your comment — I was reviewing the other PR

src/types/DataTypes.sol
```diff
@@ -126,27 +126,21 @@ library MerkleLT {
...
-        IERC20 token;
-    }
-
-    /// @notice Struct encapsulating the vesting start time and end time.
-    /// @param startTime The timestamp when the airdrop token begins to unlock.
-    /// @param endTime The timestamp when the airdrop token unlocks 100%.
-    struct Schedule {
         uint40 startTime;
-        uint40 endTime;
+        IERC20 token;
```
PaulRBerg, line 144: Does the order of declaration really matter? It's an in-memory struct

tests/Base.t.sol
```diff
@@ -542,9 +542,10 @@ abstract contract Base_Test is Assertions, Constants, DeployOptimized, Merkle, F
         return computeMerkleVCAAddress(
             merkleVCAConstructorParams({
                 campaignCreator: users.campaignCreator,
+                endTime: RANGED_STREAM_END_TIME,
                 expiration: EXPIRATION,
                 merkleRoot: MERKLE_ROOT,
-                schedule: merkleVCASchedule(),
+                startTime: RANGED_STREAM_START_TIME,
```
PaulRBerg, line 548: we definitely need a new constant for this because the VCA != ranged airdrop

tests/fork/merkle-campaign/MerkleVCA.t.sol
```diff
@@ -33,32 +33,35 @@ abstract contract MerkleVCA_Fork_Test is MerkleBase_Fork_Test {
...
-    function testForkFuzz_MerkleVCA(Params memory params, MerkleVCA.Schedule memory schedule) external {
+    function testForkFuzz_MerkleVCA(Params memory params, uint40 endTime, uint40 startTime) external {
         /*//////////////////////////////////////////////////////////////////////////
                                           CREATE
         //////////////////////////////////////////////////////////////////////////*/
 
         preCreateCampaign(params);
 
-        vm.assume(schedule.endTime > 0 && schedule.startTime > 0);
+        vm.assume(endTime > 0 && startTime > 0);
```
PaulRBerg, line 43: Why not use `bound` here?

tests/integration/fuzz/MerkleVCA.t.sol
```diff
@@ -137,14 +144,19 @@ contract MerkleVCA_Fuzz_Test is Shared_Fuzz_Test {
...
-        uint256 claimAmount = getBlockTimestamp() < schedule.endTime
-            ? (uint256(leafData.amount) * (getBlockTimestamp() - schedule.startTime))
-                / (schedule.endTime - schedule.startTime)
-            : leafData.amount;
-        uint256 forgoneAmount = leafData.amount - claimAmount;
+            claimAmount = (uint256(leafData.amount) * elapsedTime) / totalTime;
+            forgoneAmount = leafData.amount - claimAmount;
+        } else {
+            claimAmount = leafData.amount;
+        }
```
PaulRBerg, line 159: Can we move this into a bespoke mirror function somewhere in the tests?

tests/fork/merkle-campaign/MerkleVCA.t.sol
smol-ninja: Correct!

tests/integration/fuzz/MerkleVCA.t.sol
smol-ninja: Yes!

PaulRBerg (changes requested):
changes requested below

tests/Base.t.sol
smol-ninja: Similar to `VCA_FULL_AMOUNT`, I will add `VCA_END_TIME` and `VCA_START_TIME`.

PaulRBerg (approved):
LGTM now

andreivladbrg (commented):
Good job — I just had a brief look at `src`, and it looks good.

The explanatory comments seem to have been updated to use the term "sentinel", but it was missed here:  
https://github.com/sablier-labs/airdrops/blob/05c302fd3a5feb6a9c8388b0bb1c57cfe4d27b51/src/types/DataTypes.sol#L69

Also, this should be (remove first "value") :  
`/// @dev A start time of zero is used as a sentinel value for block.timestamp.`  
https://github.com/sablier-labs/airdrops/blob/05c302fd3a5feb6a9c8388b0bb1c57cfe4d27b51/src/interfaces/ISablierMerkleLL.sol#L16

src/types/DataTypes.sol
```diff
@@ -17,6 +17,7 @@ library FactoryMerkle {
 
 library MerkleInstant {
     /// @notice Struct encapsulating the constructor parameters of Merkle Instant campaigns.
+    /// @dev The fields are arranged alphabetically.
```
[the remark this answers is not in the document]
PaulRBerg: No, it is absolutely not obvious.

typically in Solidity, structs are arranged in a haphazard way in order to save gas via tight variable packing. So my first, knee-jerk guess and reaction was that the fields were arranged in that way.

While reviewing the PR, I spent 30 minutes refactoring the code to use a more optimized order, only to then realize that the fields were arranged alphabetically because these structs are used only in-memory.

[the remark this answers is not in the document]
andreivladbrg: It says "constructor params", which clearly points out it's about the "memory" struct, IMO.

I'm ok with the comment — not against it. I guess I just found it a bit funny. For me, it's like pointing with two fingers at something instead of just one :)) (metaphor)

Issue #119, opened 2025-03-23
Refactor MerkleLL "schedule" into standalone getters


PaulRBerg:
As discussed [here](https://github.com/sablier-labs/airdrops/issues/106#issuecomment-2741782757) and similar to [this issue](https://github.com/sablier-labs/airdrops/issues/111), let's refactor the `schedule` struct into standalone getters.

Doing so should also save a little gas in the claim function since loading basic types should be more gas efficient than loading a struct.

Issue #106, opened 2025-03-20
Rename "timestamps" to "schedule" in VCA


PaulRBerg:
For consistency with the MerkleLL contract, we could rename the `timestamps` to `getSchedule`:

https://github.com/sablier-labs/airdrops/blob/41f2e742744caabd64c388a32aa6eaf26bbe4654/src/interfaces/ISablierMerkleVCA.sol#L26

smol-ninja:
MerkleLL's Schedule is much more than MerkVCA's timestamps.

#### MerkleLL's Schedule

```solidity
struct Schedule {
    uint40 startTime;
    UD2x18 startPercentage;
    uint40 cliffDuration;
    UD2x18 cliffPercentage;
    uint40 totalDuration;
}
```

#### MerkVCA's timestamps

```solidity
struct Timestamps {
    uint40 start;
    uint40 end;
}
```

We initially called it `schedule` for consistency but then came to realize that maximum differentiation would be helpful here, to avoid assuming that the two represent the same struct.

PaulRBerg:
I suggest using `MerkleVCA.Schedule`

- The timestamps in `Lockup` are much more than the timestamps in `MerkleVCA`
- The `claim` function in `MerkleInstant` does a very different thing compared to `MerkleLL`, `MerkleLT`, and `MerkleVCA` — so if we were to apply your logic, we should rename the claim functions to `claimInstant`, etc.
- It'd be strange to use `Timestamps` in `Lockup` and in `MerkleVCA`, but use `Schedule` for `MerkleLL`. It'd be waaay cleaner for all Merkle contracts to use `Schedule`.
- The `Schedule` struct is always prefixed by the `MerkleVCA` library.
- 'Schedule' is actually a better fit for `MerkleVCA.Schedule` rather than `MerkleLL.Schedule` because a schedule implies an end time but that struct only contains the seeding parameters from which a schedule can be derived. Thus, a more fitting name for `MerkleLL.Schedule` would have been `Vesting` or `VestingParams` or something involving the word 'vesting'.

smol-ninja:
My argument that maximum differentiation should be achieved by using a different name for struct might be controversial given that it will always be prefixed. However, I think your point (4) contradicts the argument made in point (3).

Using points (1), (2) and (3), you are establishing an argument that prefixing `Schedule` in `MerkleLL` and `MerkleVCA` avoids confusion even though they represent different structs. So, the prefix is what provides differentiation over the struct name.

[...]

PR #125, opened 2025-03-25
refactor: merkleLL schedule into immutable vars

smol-ninja:
Closes https://github.com/sablier-labs/airdrops/issues/119.

The `lite` profile cannot compile the code without optimizer due to stack issue. So, I have removed it. Also, now the variables are immutable, it would be much more efficient to load them as they are no longer a part of the contract storage.

### Note

Code coverage CI depends on https://github.com/sablier-labs/reusable-workflows/pull/21


andreivladbrg (commented):
IMO we should keep the `Schedule` struct to group together the params related to `MerkleLL`.

We can keep the internal vars as `immutable` and `private`, and retrieve them through the `getSchedule` getter:


```solidity
uint40 private immutable CLIFF_DURATION;
UD60x18 private immutable CLIFF_UNLOCK_PERCENTAGE;
uint40 private immutable START_TIME;
UD60x18 private immutable START_UNLOCK_PERCENTAGE;
uint40 private immutable TOTAL_DURATION;

function getSchedule() public view returns (MerkleLL.Schedule memory) {
    return MerkleLL.Schedule({
        cliffDuration: CLIFF_DURATION,
        cliffUnlockPercentage: CLIFF_UNLOCK_PERCENTAGE,
        startTime: START_TIME,
        startUnlockPercentage: START_UNLOCK_PERCENTAGE,
        totalDuration: TOTAL_DURATION
    });
}
```

*(IIRC, reading private constants is slightly more gas-efficient than reading public ones — but just slightly)*

Also, similar [to this](https://github.com/sablier-labs/airdrops/pull/128#discussion_r2012486474):
Since these are now constant variables (part of bytecode), there’s no point in trying to group them into the same storage slot, as there are no slots.

andreivladbrg:
> What is the benefit? Razvan mentioned https://github.com/sablier-labs/airdrops/issues/106#issuecomment-2743436134 that they don't use it in subgraphs so it wont break anything anyways.

I find this better structured. And if gas efficiency is important, it actually saves even more gas for `_claim`.  
When you declare a variable as `public`, Solidity automatically generates a getter function for it. This means that accessing the variable externally requires executing that function, which adds some gas overhead.

> Yes they are not grouped but follow alphabetical order

I don’t think you got my point — it’s not about the order, but about the type: `UD2x18` vs `UD60x18`.
We originally used `UD2x18` to group the variables together (gas efficient), but now there’s nothing to group (constants).

smol-ninja:
1. Having public immutable has less contract size than having private + accessing via a function returning struct.

2. Accessing private immutable is more gas efficient than accessing public immutable. You were right @andreivladbrg, I was wrong.

However, I tested with 3 immutable variables, in case of (1) the gas increase in deployment is 8k and in case of (2) gas reduction in claim is only 44. So, I still think having public immutable is better. Wdyt?

andreivladbrg:
Thanks you for the in-depth research.  
You're absolutely right that using `private` + getters increases the bytecode. You're also right that using a `struct` may increase it even more.  
I must admit I hadn't considered these implications in my original message.

Now, regarding the decision-making: is having greater bytecode (at this point) problematic? I would say no, since we now have different factories.  
As for deployment gas costs, yes, they may increase — but that’s a one-time cost paid by the creator, who is (presumably) well-funded.  
In the case of casting, it would be done as `recipient_count * claim`, which—even if it's a small gas cost per instance—could become more expensive overall if there’s a large number of recipients.

> I find UD2x18 more useful over UD60x18 to represent percentages whereas UD60x18 is more useful to represent amounts. 

I believe this perception might come from the fact that we often pack variables using `uint64` to save on storage gas.  
I don't think there’s a strong correlation between a specific PRBMath type and its usage context. (we had in the past pct values as `UD60x18`)
The only exception might be time variables, for which we consistently use `uint40` regardless.

> Given that both would occupy 32 bytes being immutable, so no gas difference. Plus, using ud2x18 in constructor which mean less calldata (slight). For these reasons, I think ud2x18 is also not bad in this context.

Apologies, I didn’t quite understand this point — why is calldata size relevant in this context?  
The bytecode itself shouldn’t be affected, as the actual numbers are ABI-encoded and appended at the end of the bytecode.

---

So, my current conclusion is that we should go with `private` constants + getters  using `UD60x18` for percentage values, as this would provide the best claim experience. Also, no more need for a `Schedule` struct now from my perspective.

smol-ninja:
> is having greater bytecode (at this point) problematic? I would say no, since we now have different factories

I was referring to merkle contracts and not factories. But I think you got the point.

> I believe this perception might come from the fact that we often pack variables using uint64 to save on storage gas.

The perception is from seeing `UD2x18` as a number with 2 bits for the integer and 18 bits for fraction (similar to percentage format). For example, see [Q format](https://en.wikipedia.org/wiki/Q_(number_format)) where `Q8.8` means 8 integer bits and 8 fraction bits.

> why is calldata size relevant in this context

Since we were comparing gas differences (even though small), a calldata with `ud2x18` values will have less size than calldata with `UD60x18` values. Campaign creators pay for calldata.

> my current conclusion is that we should go with private constants + getters 

Wouldn't it require us to follow the same practice throughout our protocols, including mutable variables? For ex. the same argument applies for [`nextStreamId`](https://github.com/sablier-labs/lockup/blob/main/src/abstracts/SablierLockupBase.sol#L39) as well. If accessing private constants is efficient, then accessing private mutable variables should also be efficient. Wdyt?

andreivladbrg:
> I was referring to merkle contracts and not factories. But I think you got the point.

Yes, I got it — I mentioned the factories because previously we only had a factory contract (which included the campaigns' bytecode). That contract exceeded the size limit, so worrying about bytecode size was indeed relevant back then. But now, since there’s a large margin, it's fine.

image


> The perception is from seeing UD2x18 as a number with 2 bits for the integer and 18 bits for fraction (similar to percentage format). For example, see [Q format](https://en.wikipedia.org/wiki/Q_(number_format)) where Q8.8 means 8 integer bits and 8 fraction bits.

Thanks for the explanation. To be honest, I'm not sure why having 2 bits or 60 bits for the integer part should make a difference — or if one communicates that `100% = 1e18` more clearly than the other.

> Since we were comparing gas differences (even though small), a calldata with ud2x18 values will have less size than calldata with UD60x18 values. Campaign creators pay for calldata. Anyways you can ignore it for now as its a tiny deal

First of all, what matters the most is the `claim` gas.  
Second, campaign creators would also pay for casting `UD2x18` to `UD60x18`, so I’m not sure there would be a real improvement. But I agree with “it's a tiny deal” — again, what really matters is the `claim`.

> Wouldn't it require us to follow the same practice throughout our protocols, including mutable variables? For ex. the same argument applies for [nextStreamId](https://github.com/sablier-labs/lockup/blob/main/src/abstracts/SablierLockupBase.sol#L39) as well. If accessing private constants is efficient, then accessing private mutable variables should also be efficient. Wdyt

You’re right.  
Though in the case of `Lockup`, we’re already at the contract size limit — not sure if these kinds of optimizations would help us much.

smol-ninja:
> To be honest, I'm not sure why having 2 bits or 60 bits for the integer part should make a difference

I am convinced. I will change it to `UD60x18`.

> Re. visibility of immutable variables

I am not convinced yet. I don't think it would have much impact on contract size of Lockup either. We have public immutable and mutable variables in `SablierMerkleBase.sol`, Flow etc. as well. So, if were to focus on this, we should discuss it in a separate discussion. More than constants, it affects storage variables of standard types and yet we use `public` visibility for storage variables because they are handy, and gas diff is slightly. So my take is if we want to make them private, why just do it in `SablierMerkleLL`, it should be done throughout the code. 

I also think its a problem of solidity not EVM so we can start a discussion in their repo as well.

andreivladbrg:
Alright. Let's go for public `UD60x18` - thats ok

smol-ninja:
Don't mind the failing fork test (in case it fails). Thats because the recipient could be the campaign contract and that fucks up with the ETH admin balance check. I have already fixed it in https://github.com/sablier-labs/airdrops/pull/128.

tests/integration/concrete/campaign/ll/claim/claim.t.sol
```diff
@@ -1,8 +1,7 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity >=0.8.22 <0.9.0;
 
-import { ud2x18 } from "@prb/math/src/UD2x18.sol";
-import { ud60x18 } from "@prb/math/src/UD60x18.sol";
+import { ud, ud60x18, ZERO } from "@prb/math/src/UD60x18.sol";
```
andreivladbrg, line 4: both `ud` and `ud60x18` imported, only one should suffice

andreivladbrg (commented):
> The lite profile cannot compile the code without optimizer due to stack issue. So, I have removed it. Also, now the

the reason for the `lite` profile wasn’t strictly the `optimizer = false`, but the fuzz runs: 50 vs. 1000. i’d say keep it, so we can just turn the optimizer on — not a big change

PaulRBerg:
- Re `UD60x18` vs `UD2x18` — it doesn't really matter. Happy to use `UD60x18`. Andrei is right that there's no association between `UD2x18` and percentages. We used to use `UD2x18` to save gas via tight variable packing.
- Let's use public immutable vars. As Shub said, that's how we typically represent immutables in other Sablier contracts. If you Andrei want to propose a new model for all of our immutable vars, please open a separate discussion.
    - Note: I don't think that that would be a good idea. That refactor would increase the diff between our releases, which would increase our audit costs.

> I'm not sure why having 2 bits or 60 bits for the integer part should make a difference

Note: in PRBMath, the numbers refer to digits, so that's 2 digits and 60 digits.

PaulRBerg (approved):
LGTM after my commit

src/interfaces/ISablierMerkleLL.sol
```diff
@@ -11,8 +12,20 @@ interface ISablierMerkleLL is ISablierMerkleLockup {
...
-    function getSchedule() external view returns (MerkleLL.Schedule memory);
+    ///@notice Returns the duration of the cliff in seconds.
+    function CLIFF_DURATION() external view returns (uint40);
+
+    /// @notice Returns the percentage of the claim amount to be unlocked at cliff time, as a fixed-point number where
+    /// 1e18 is 100%.
+    function CLIFF_UNLOCK_PERCENTAGE() external view returns (UD60x18);
+
+    /// @notice Returns the start time of the stream. Zero is a sentinel value for `block.timestamp`.
+    function START_TIME() external view returns (uint40);
```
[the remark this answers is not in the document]
smol-ninja: 1. The campaign does not have a definition of start time. The start of the campaign is restricted by its funding status.
2. Having said that, since it cannot be confused, should we still rename it?

src/SablierMerkleLL.sol
```diff
@@ -42,19 +42,19 @@ contract SablierMerkleLL is
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @inheritdoc ISablierMerkleLL
-    uint40 public immutable override CLIFF_DURATION;
+    uint40 public immutable override VESTING_CLIFF_DURATION;
```
smol-ninja, line 45: I think this is incorrect terminology. For instance, see the below screenshot:
Screenshot 2025-03-26 at 16 45 18

Technically, vesting is what begins after cliff. So `VESTING_TOTAL_DURATION` can be confused with total duration minus cliff duration.

I don't think prefixing with VESTING is any helpful.

PaulRBerg: If `2 year vesting` means 3 years in total in that image you shared, that is an appalling use of language in the contract from which you have taken that screenshot.

Normally, the total vesting duration = cliff + vesting after cliff

ChatGPT 4o, ChatGPT o1, and Claude Sonnet 3.7 all agree:


SCR-20250326-qlzv
SCR-20250326-qmfv

src/interfaces/ISablierMerkleLL.sol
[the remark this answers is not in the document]
PaulRBerg: A reader of the contract doesn't know that the campaign doesn't have a start time. You know that, but you assume that everyone would know that, too.

We already have `STREAM_` prefixes for the `MerkleLockup` variables ,so I think that this `VESTING_` prefix comes in handy.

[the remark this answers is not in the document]
PaulRBerg: `START_TIME` could be misunderstood for _time when claims are open_.

Whereas `VESTING_START_TIME` is much harder to be misunderstood for that.

andreivladbrg:
> Re UD60x18 vs UD2x18 — it doesn't really matter it does 

It does matter in terms of gas, as it would be one less cast (from `UD2x18` to `UD60x18` — when calculating the unlock pcts) — saving gas during the claim

PaulRBerg:
@andreivladbrg yes but the difference is tiny (44 gas), so we can nevermind it.

src/SablierMerkleLL.sol
smol-ninja: OK if ChatGPT says so, he knows more than me when it comes to human language :))

src/interfaces/ISablierMerkleLL.sol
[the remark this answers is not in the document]
smol-ninja: Sounds good.

smol-ninja:
I am resolving the conflicts on this PR.

PaulRBerg:
@smol-ninja just resolved the conflicts now haha

tests/integration/concrete/campaign/ll/constructor.t.sol
```diff
@@ -1,46 +1,6 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity >=0.8.22 <0.9.0;
 
+import { console2 } from "forge-std/src/console2.sol";
 import { SablierMerkleLL } from "src/SablierMerkleLL.sol";
-import { MerkleLL } from "src/types/DataTypes.sol";
-
 import { Integration_Test } from "./../../../Integration.t.sol";
-
-contract Constructor_MerkleLL_Integration_Test is Integration_Test {
```
smol-ninja, line 9: This got deleted accidentally.

smol-ninja: `bun run lint` caught it.

smol-ninja:
@andreivladbrg waiting for your approval.

andreivladbrg:
@smol-ninja not sure why the coverage has decreased

smol-ninja:
> why the coverage has decreased

I don't know but I will look into it later. This might be because of `via-ir-minimum`.

Issue #130, opened 2025-03-26
Add "VESTING_" prefix in "MerkleLT" and "MerkleVCA"


PaulRBerg:
Context: https://github.com/sablier-labs/airdrops/pull/125

- [x] Rename `END_TIME` to `VESTING_END_TIME`
- [x] Rename `STREAM_START_TIME`  to `VESTING_START_TIME`
    - Rename `params.streamStartTime` to `params.startTime` (for consistency)
- [x] Rename `START_TIME` to `VESTING_START_TIME`

PaulRBerg:
We just need to tackle this and we can hand off the final commits

PR #133, opened 2025-03-27
refactor: prefix timestamps with VESTING in MerkleLT and MerkleVCA

smol-ninja:
Closes https://github.com/sablier-labs/airdrops/issues/130

### DIFF
--- a/src/SablierMerkleVCA.sol
+++ b/src/SablierMerkleVCA.sol
@@ -99,14 +99,10 @@ contract SablierMerkleVCA is
             revert Errors.SablierMerkleVCA_UnlockPercentageTooHigh(params.unlockPercentage);
         }
 
-        // Effect: set the vesting end time.
-        VESTING_END_TIME = params.endTime;
-
-        // Effect: set the vesting start time.
-        VESTING_START_TIME = params.startTime;
-
-        // Effect: set the unlock percentage.
+        // Effect: set the immutable variables.
         UNLOCK_PERCENTAGE = params.unlockPercentage;
+        VESTING_END_TIME = params.endTime;
+        VESTING_START_TIME = params.startTime;
     }
 
     /*//////////////////////////////////////////////////////////////////////////
--- a/src/SablierMerkleVCA.sol
+++ b/src/SablierMerkleVCA.sol
@@ -41,13 +41,13 @@ contract SablierMerkleVCA is
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @inheritdoc ISablierMerkleVCA
-    uint40 public immutable override END_TIME;
-
-    /// @inheritdoc ISablierMerkleVCA
-    uint40 public immutable override START_TIME;
-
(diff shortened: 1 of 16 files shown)


src/types/DataTypes.sol
```diff
@@ -103,7 +104,7 @@ library MerkleLT {
         ISablierLockup lockup;
         bytes32 merkleRoot;
         string shape;
-        uint40 streamStartTime;
+        uint40 startTime;
```
andreivladbrg, line 107: why not `vestingStartTime`?

same q on `MerkleLT`

src/types/DataTypes.sol
```diff
@@ -131,7 +132,7 @@ library MerkleVCA {
     /// @param initialAdmin The initial admin of the campaign.
     /// @param ipfsCID The content identifier for indexing the contract on IPFS.
     /// @param merkleRoot The Merkle root of the claim data.
-    /// @param startTime Vesting start time, as a Unix timestamp. Zero is a sentinel value for `block.timestamp`.
+    /// @param startTime Vesting start time, as a Unix timestamp.
```
andreivladbrg, line 135: good catch 👍

andreivladbrg (approved):
Generally looks good. Just some questions:

This comment: `// Zero is a sentinel value for block.timestamp.` appears in both `ISablierMerkleVCA.calculateClaimAmount` and the actual implementation.  
Same for `calculateForgoneAmount` — should we keep it only in the interface, given that we have `/// @inheritdoc ISablierMerkleVCA`?

src/SablierMerkleVCA.sol
```diff
@@ -99,11 +99,11 @@ contract SablierMerkleVCA is
...
 
-        // Effect: set the end time.
-        END_TIME = params.endTime;
+        // Effect: set the vesting end time.
+        VESTING_END_TIME = params.endTime;
 
-        // Effect: set the start time.
-        START_TIME = params.startTime;
+        // Effect: set the vesting start time.
+        VESTING_START_TIME = params.startTime;
```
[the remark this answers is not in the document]
andreivladbrg: 31f4153e1aab13460948d2e001dbd5166ae3b462

smol-ninja:
> This comment: `// Zero is a sentinel value for block.timestamp.` appears in both `ISablierMerkleVCA.calculateClaimAmount` and the actual implementation.  
Same for `calculateForgoneAmount` — should we keep it only in the interface, given that we have `/// @inheritdoc ISablierMerkleVCA`?

Yes that's deliberate because its important.

src/types/DataTypes.sol
smol-ninja: It is mentioned in the [issue description](https://github.com/sablier-labs/airdrops/issues/130).

andreivladbrg:
> Yes that's deliberate because its important

ok, agree

PaulRBerg (commented):
this is merged but _it would have looked good to me_ 😉

PR #137, opened 2025-03-27
docs: natspec retouches

PaulRBerg:
For https://github.com/sablier-labs/airdrops/pull/133

### DIFF
--- a/src/interfaces/ISablierMerkleLL.sol
+++ b/src/interfaces/ISablierMerkleLL.sol
@@ -19,7 +19,8 @@ interface ISablierMerkleLL is ISablierMerkleLockup {
     /// fixed-point number where 1e18 is 100%.
     function VESTING_CLIFF_UNLOCK_PERCENTAGE() external view returns (UD60x18);
 
-    /// @notice Retrieves the start time of the vesting stream. Zero is a sentinel value for `block.timestamp`.
+    /// @notice Retrieves the start time of the vesting stream, as a Unix timestamp. Zero is a sentinel value for
+    /// `block.timestamp`.
     function VESTING_START_TIME() external view returns (uint40);
 
     /// @notice Retrieves the percentage of the claim amount due to be unlocked at the vesting start time, as a
--- a/src/interfaces/ISablierMerkleVCA.sol
+++ b/src/interfaces/ISablierMerkleVCA.sol
@@ -24,22 +24,25 @@ interface ISablierMerkleVCA is ISablierMerkleBase {
     /// value is denominated as a fixed-point number where 1e18 is 100%.
     function UNLOCK_PERCENTAGE() external view returns (UD60x18);
 
-    /// @notice Retrieves the timestamp when the airdrop token unlocks 100%.
+    /// @notice Retrieves the time when the VCA airdrop is fully vested, as a Unix timestamp.
     function VESTING_END_TIME() external view returns (uint40);
 
-    /// @notice Retrieves the timestamp when the airdrop token begins to unlock.
+    /// @notice Retrieves the time when the VCA airdrop begins to unlock, as a Unix timestamp. Zero is a sentinel value
+    /// for `block.timestamp`.
     function VESTING_START_TIME() external view returns (uint40);
 
     /// @notice Calculates the amount that would be claimed if the claim were made at `claimTime`.
     /// @dev This is for informational purposes only. To actually claim the airdrop, a Merkle proof is required.
     /// @param fullAmount The amount of tokens allocated to a user, denominated in the token's decimals.
     /// @param claimTime A hypothetical time at which to make the claim. Zero is a sentinel value for `block.timestamp`.
+    /// @return The amount that would be claimed, denominated in the token's decimals.
     function calculateClaimAmount(uint128 fullAmount, uint40 claimTime) external view returns (uint128);
 
     /// @notice Calculates the amount that would be forgone if the claim were made at `claimTime`.
     /// @dev This is for informational purposes only. To actually claim the airdrop, a Merkle proof is required.
     /// @param fullAmount The amount of tokens allocated to a user, denominated in the token's decimals.
     /// @param claimTime A hypothetical time at which to make the claim. Zero is a sentinel value for `block.timestamp`.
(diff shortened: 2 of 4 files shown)


src/interfaces/ISablierMerkleLL.sol
```diff
@@ -19,7 +19,8 @@ interface ISablierMerkleLL is ISablierMerkleLockup {
     /// fixed-point number where 1e18 is 100%.
     function VESTING_CLIFF_UNLOCK_PERCENTAGE() external view returns (UD60x18);
 
-    /// @notice Retrieves the start time of the vesting stream. Zero is a sentinel value for `block.timestamp`.
+    /// @notice Retrieves the start time of the vesting stream, as a Unix timestamp. Zero is a sentinel value for
```
smol-ninja, line 22: Just a question, if we don't mention `as a Unix timestamp` explicitly, what else can it be confused with? Isn't mentioning `time` enough? All times are unix timestamps in the context of programming language.

src/interfaces/ISablierMerkleVCA.sol
```diff
@@ -24,22 +24,25 @@ interface ISablierMerkleVCA is ISablierMerkleBase {
     /// value is denominated as a fixed-point number where 1e18 is 100%.
     function UNLOCK_PERCENTAGE() external view returns (UD60x18);
 
-    /// @notice Retrieves the timestamp when the airdrop token unlocks 100%.
+    /// @notice Retrieves the time when the VCA airdrop is fully vested, as a Unix timestamp.
     function VESTING_END_TIME() external view returns (uint40);
 
-    /// @notice Retrieves the timestamp when the airdrop token begins to unlock.
+    /// @notice Retrieves the time when the VCA airdrop begins to unlock, as a Unix timestamp. Zero is a sentinel value
```
smol-ninja, line 30: No, zero is not a sentinel value in MerkleVCA.

smol-ninja (changes requested):

src/interfaces/ISablierMerkleLL.sol
PaulRBerg: When I went to Thailand, I was surprised to see my Google results show up in the year 2567.

Plus, this isn't quite the same issue, but [time & timezones are notorously problematic](https://youtu.be/-5wpm-gesOY?si=zo42rYiBlP4cfkCO) in computer science.

src/interfaces/ISablierMerkleVCA.sol
PaulRBerg: Good catch!

PaulRBerg: Feel free to push and merge

src/interfaces/ISablierMerkleLL.sol
smol-ninja: >  I was surprised to see my Google results show up in the year 2567.

Thats the issue with converting unix timestamp into readable date format. Under the hood, it would have returned the correct unix timestamp. The conversion to date format is a different issue.

> but [time & timezones are notorously problematic](https://youtu.be/-5wpm-gesOY?si=zo42rYiBlP4cfkCO) in computer

Yes they are but UNIX timestamp is universal and does not change with timezones.

smol-ninja: Plus the date format is a string format. So, if the function returns `uint60`, it can definitely not be confused with the date format. For that, it has to return a `string`.

src/interfaces/ISablierMerkleVCA.sol
smol-ninja: https://github.com/sablier-labs/airdrops/pull/137/commits/627dfcaec09eff3ede42cf814b57e0b831ac77bc

src/interfaces/ISablierMerkleLL.sol
PaulRBerg: How do you know that `startTime` is a UNIX timestamp without the comment that says that it is a UNIX timestamp?

Even if we exclude dates (which should be `string`), the time could be confused for:

- A date of the format `YYYYMMDD`, e.g. `20250328` (this is still an integer!)
- An integer/hexadecimal representation of a string
- An hour or a day

Now, I know that these confusions are unlikely. But it doesn't cost us much to add that comment there.

It is also worth noting that:

1. **We're still early stage.** There's like [24k](https://www.developerreport.com/) monthly active developers working in crypto. Among the next 100k developers to join crypto, there may be newbies who have no idea what UNIX timestamps are, and that they are the default time unit in computing.
2. **The blockchain is like amber.** The immutability of the blockchain guarantees that our work will survive for many decades, and potentially hundreds of years ahead. Then, UNIX timestamps might be an archaic concept. However, English will most likely still be in use, and the English-based explanation for the unit used for `startTime` will be super helpful to these smart contract readers from the future.

smol-ninja: > How do you know that startTime is a UNIX timestamp without the comment that says that it is a UNIX timestamp?

By looking at the return type - an integer. Etherscan clearly shows the return type. And for someone looking at the code, knowing that it returns an integer implies that it's the timestamp. Given UNIX timestamp is the only timestamp that exists as an integer, it implies it's a UNIX timestamp.

> the time could be confused for `20250328`, integer representation of a string, An hour or a day.

Haha, yeah, that's possible.

> it doesn't cost us much to add that comment there

I know, I know. I was just considering it from the perspective of whether it's helpful at all. I am in agreement that comments are helpful, even if the value addition is marginally small. But in this context, I thought the value add is 0..

> there may be newbies who have no idea what UNIX timestamps

Hmm, good point about teaching it to someone who has no idea about handling timestamps in computers. Definitely value add for them.

> The immutability of the blockchain guarantees that our work will survive for many decades, and potentially hundreds of years ahead

Perfect point.

I am in agreement now. The reason I wanted to discuss it was so I can be better at writing comments myself instead of you correcting them afterward, which requires me to understand the thought process you had behind all these.

PaulRBerg: Thanks

I enjoy these discussions, keep the coming!

PR #140, opened 2025-04-02
refactor: remove superflous check in _calculateClaimAmount

andreivladbrg:
Addresses the points from here: https://cantina.xyz/code/c9980035-706a-4119-91fe-7ebddaf123e4/findings?finding=13


src/SablierMerkleInstant.sol
```diff
@@ -64,6 +64,6 @@ contract SablierMerkleInstant is
         TOKEN.safeTransfer(recipient, amount);
 
         // Log the claim.
-        emit Claim(index, recipient, amount);
```
smol-ninja, line 67: I just sent my comment on the audit dashboard but this was changed deliberately because the named args are same as values. Its just duplication. 

If you agree, we should roll it back.

smol-ninja (changes requested):
You haven't handled the check in `_calculateClaimAmount`. What about 1st and 3rd finding?

src/SablierMerkleInstant.sol
andreivladbrg: > If you agree, we should roll it back.

i have no preference (thats why i changed it). the pro is that its consistent with all events - the cons is what you said, duplication

once Akshay respondes, feel free to bring back

smol-ninja: Sometimes a bit of an inconsistency can be better :))

Will wait for Akshay's reply.

SECURITY.md
```diff
@@ -21,3 +21,6 @@ vulnerability, it must adhere to these assumptions as well:
 - The token contract has only one entry point.
 - The token contract does not allow callbacks (e.g. ERC-777 is not supported).
 - The native token of the chain, i.e. the one used for `msg.value` payments, has 18 decimals.
+- Do not pre-fund an airdrop campaign (i.e., sending tokens to the to-be-deployed address) before creating it through
+  the factory. Otherwise, if the provided parameters are invalid, the campaign will fail to be created, and the funds
```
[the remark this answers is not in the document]
andreivladbrg: ca72acf128b15494c43319fa415010b6c4216444

src/SablierMerkleInstant.sol
smol-ninja: Closing it as it does not matter.

smol-ninja (approved):

PR #142, opened 2025-04-07
feat: add fee recipient to `collectFees` function

smol-ninja:
### Changelog

- Closes https://github.com/sablier-labs/airdrops/issues/139
- Changes the type of `FACTORY` constant to `ISablierFactoryMerkleBase`. The number of times its casted from `address` to `ISablierFactoryMerkleBase` would be higher after RBAC PR. So it would be useful.

### Notes

I have used `to` as the input parameter of `collectFees` functions. However, I have referred to it as fee recipient or `feeRecipient` in tests, events and errors. I would prefer renaming `to` to `feeRecipient` in function parameter as well. Let me know if you agree.

### DIFF
--- a/src/abstracts/SablierFactoryMerkleBase.sol
+++ b/src/abstracts/SablierFactoryMerkleBase.sol
@@ -62,17 +62,17 @@ abstract contract SablierFactoryMerkleBase is
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @inheritdoc ISablierFactoryMerkleBase
-    function collectFees(ISablierMerkleBase campaign, address to) external override {
-        // Check: if `msg.sender` is not the admin, `to` must be the admin address.
-        if (msg.sender != admin && to != admin) {
-            revert Errors.sablierMerkleFactoryBase_FeeRecipientNotAdmin({ feeRecipient: to, admin: admin });
+    function collectFees(ISablierMerkleBase campaign, address feeRecipient) external override {
+        // Check: if `msg.sender` is not the admin, `feeRecipient` must be the admin address.
+        if (msg.sender != admin && feeRecipient != admin) {
+            revert Errors.SablierMerkleFactoryBase_FeeRecipientNotAdmin({ feeRecipient: feeRecipient, admin: admin });
         }
 
         // Effect: collect the fees from the campaign contract.
-        uint256 feeAmount = campaign.collectFees(to);
+        uint256 feeAmount = campaign.collectFees(feeRecipient);
 
         // Log the fee withdrawal.
-        emit CollectFees({ feeRecipient: to, campaign: campaign, feeAmount: feeAmount });
+        emit CollectFees({ admin: admin, campaign: campaign, feeRecipient: feeRecipient, feeAmount: feeAmount });
     }
 
     /// @inheritdoc ISablierFactoryMerkleBase
--- a/src/abstracts/SablierFactoryMerkleBase.sol
+++ b/src/abstracts/SablierFactoryMerkleBase.sol
@@ -62,12 +62,17 @@ abstract contract SablierFactoryMerkleBase is
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @inheritdoc ISablierFactoryMerkleBase
-    function collectFees(ISablierMerkleBase campaign) external override {
+    function collectFees(ISablierMerkleBase campaign, address to) external override {
+        // Check: if `msg.sender` is not the admin, `to` must be the admin address.
+        if (msg.sender != admin && to != admin) {
+            revert Errors.sablierMerkleFactoryBase_FeeRecipientNotAdmin({ feeRecipient: to, admin: admin });
+        }
+
         // Effect: collect the fees from the campaign contract.
-        uint256 feeAmount = campaign.collectFees(admin);
+        uint256 feeAmount = campaign.collectFees(to);
 
         // Log the fee withdrawal.
-        emit CollectFees({ admin: admin, campaign: campaign, feeAmount: feeAmount });
+        emit CollectFees({ feeRecipient: to, campaign: campaign, feeAmount: feeAmount });
     }
 
     /// @inheritdoc ISablierFactoryMerkleBase
--- a/src/interfaces/ISablierFactoryMerkleBase.sol
+++ b/src/interfaces/ISablierFactoryMerkleBase.sol
@@ -13,7 +13,9 @@ interface ISablierFactoryMerkleBase is IAdminable {
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @notice Emitted when the accrued fees are collected.
-    event CollectFees(address indexed feeRecipient, ISablierMerkleBase indexed campaign, uint256 feeAmount);
+    event CollectFees(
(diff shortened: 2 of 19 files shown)


src/abstracts/SablierFactoryMerkleBase.sol
```diff
@@ -62,12 +62,17 @@ abstract contract SablierFactoryMerkleBase is
...
+            revert Errors.sablierMerkleFactoryBase_FeeRecipientNotAdmin({ feeRecipient: to, admin: admin });
+        }
+
         // Effect: collect the fees from the campaign contract.
-        uint256 feeAmount = campaign.collectFees(admin);
+        uint256 feeAmount = campaign.collectFees(to);
 
         // Log the fee withdrawal.
-        emit CollectFees({ admin: admin, campaign: campaign, feeAmount: feeAmount });
+        emit CollectFees({ feeRecipient: to, campaign: campaign, feeAmount: feeAmount });
```
PaulRBerg, line 75: We still want to log the admin. The fee recipient should be the 3rd param.

src/interfaces/ISablierFactoryMerkleBase.sol
```diff
@@ -13,7 +13,7 @@ interface ISablierFactoryMerkleBase is IAdminable {
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @notice Emitted when the accrued fees are collected.
-    event CollectFees(address indexed admin, ISablierMerkleBase indexed campaign, uint256 feeAmount);
+    event CollectFees(address indexed feeRecipient, ISablierMerkleBase indexed campaign, uint256 feeAmount);
```
PaulRBerg, line 16: Fee recipient should be 2nd param

src/interfaces/ISablierFactoryMerkleBase.sol
```diff
@@ -64,14 +64,19 @@ interface ISablierFactoryMerkleBase is IAdminable {
...
     //////////////////////////////////////////////////////////////////////////*/
 
-    /// @notice Collects the fees accrued in the given campaign contract, and transfers them to the factory admin.
+    /// @notice Collects the fees accrued in the given campaign contract.
     /// @dev Emits a {CollectFees} event.
     ///
     /// Notes:
-    /// - If the admin is a contract, it must be able to receive native token payments, e.g., ETH for Ethereum Mainnet.
+    /// - If the `to` address is a contract, it must be able to receive native token payments, e.g., ETH for Ethereum
+    /// Mainnet.
```
PaulRBerg, line 72: Can be moved up

src/interfaces/ISablierFactoryMerkleBase.sol
```diff
@@ -64,14 +64,19 @@ interface ISablierFactoryMerkleBase is IAdminable {
...
+    ///
+    /// Requirements:
+    /// - If `msg.sender` is not the admin, `to` must be the admin address.
     ///
     /// @param campaign The address of the Merkle contract to collect the fees from.
-    function collectFees(ISablierMerkleBase campaign) external;
+    /// @param to The address where the fees will be transferred.
```
PaulRBerg, line 78: "where the fees will be collected"

src/interfaces/ISablierMerkleBase.sol
```diff
@@ -113,21 +114,21 @@ interface ISablierMerkleBase is IAdminable {
...
     function clawback(address to, uint128 amount) external;
 
-    /// @notice Collects the accrued fees by transferring them to {FACTORY}.
+    /// @notice Collects the accrued fees by transferring them to the `to` address.
     ///
     /// Requirements:
     /// - `msg.sender` must be {FACTORY}.
     ///
-    /// @param factoryAdmin The address of the admin of {FACTORY}.
+    /// @param to The address to receive the fees.
```
PaulRBerg, line 122: "the address where the fees will be collected"

src/interfaces/ISablierMerkleBase.sol
```diff
@@ -113,21 +114,21 @@ interface ISablierMerkleBase is IAdminable {
...
-    function collectFees(address factoryAdmin) external returns (uint256 feeAmount);
+    function collectFees(address to) external returns (uint256 feeAmount);
 
     /// @notice Lowers the min USD fee.
     ///
     /// @dev Emits a {LowerMinFeeUSD} event.
     ///
     /// Requirements:
-    /// - `msg.sender` must be the admin of {FACTORY}.
+    /// - `msg.sender` must have {FEE_MANAGEMENT_ROLE} role in {FACTORY}.
```
PaulRBerg, line 131: RBAC doesn't exist yet

src/libraries/Errors.sol
```diff
@@ -42,6 +42,10 @@ library Errors {
                             SABLIER-MERKLE-FACTORY-BASE
     //////////////////////////////////////////////////////////////////////////*/
 
+    /// @notice Thrown when non-admin address calls {collectFees} without setting the fee recipient to the admin
```
PaulRBerg, line 45: "Thrown when an unauthorized address attempts to collect the fees."

src/libraries/Errors.sol
```diff
@@ -42,6 +42,10 @@ library Errors {
...
+    /// address.
+    error sablierMerkleFactoryBase_FeeRecipientNotAdmin(address feeRecipient, address admin);
```
PaulRBerg, line 47: Error name should be uppercase

PaulRBerg (changes requested):
Feedback below

Happy to rename the param to `feeRecipient`.

src/interfaces/ISablierFactoryMerkleBase.sol
smol-ninja: Caller, campaign, feeRecipient, feeAmount.

These values in this order it should emit right?

src/interfaces/ISablierMerkleBase.sol
smol-ninja: I thought I removed it from everywhere. But missed this one.

src/libraries/Errors.sol
smol-ninja: There is no unauthorised address from the perspective of collect fee. So i don't think it would be accurate to say that.

It only reverts if caller and fee recipient both are not admin.

PaulRBerg: I think if's fine to use "unauthorized" in this circumstance, but I won't belabor this point.

happy to use your version

src/abstracts/SablierFactoryMerkleBase.sol
PaulRBerg: Nah, just the admin. The caller can be obtained from the raw tx data

src/interfaces/ISablierFactoryMerkleBase.sol
smol-ninja:
```solidity
event CollectFees(
        address indexed admin, ISablierMerkleBase indexed campaign, address indexed feeRecipient, uint256 feeAmount
    );
```

src/interfaces/ISablierFactoryMerkleBase.sol
smol-ninja: I have merged it into `@notice`.

Issue #136, opened 2025-03-27
Rename "getTranchesWithPercentages" to "tranchesWithPercentages"


PaulRBerg:
So that there's no more `get` read-only function in this repo

smol-ninja:
What's the issue with `getTranchesWithPercentages`? Its a getter thats why it has `get` in its name. 

Are you suggesting to never use `get` for any function in any of the repo?

PaulRBerg:
Yeah well I can't find any good criterion for what should use `get` and what shouldn't.

So it's not an objective decion-making process, and we get bogged down in discussions about what should have 'get' as a prefix, what shouldn't, etc.

Easier to just not have that problem.

smol-ninja:
I had the same question once and @andreivladbrg answered that `get` should be used with function that returns state value unmodified. That seems like a clear rule.

If we rename `getTranchesWithPercentages` to `tranchesWithPercentages` then we break that rule and go into the discussion of what should use `get` and what should not.

So either we don't use `get` at all (read only functions have `view` modifier so it implies they return something and therefore `get` doesn't add much value imo) or we can stick to Andrei's rule. What do you think?

andreivladbrg:
I’d suggest not using these names, as they might create confusion — either suggesting there are two different variables, or that `tranchesWithPercentages` involves some additional logic that modifies its value.

```solidity
MerkleLT.TrancheWithPercentage[] internal _tranchesWithPercentages;
function tranchesWithPercentages() external view returns (MerkleLT.TrancheWithPercentage[] memory) {
     return _tranchesWithPercentages;
}
``` 

Using `get` makes it very clear that the function simply returns a value, without any additional logic.

[GPT agrees](https://chatgpt.com/share/67e68965-48e4-8013-b932-49d3bcfe4ae1).

PaulRBerg:
Your rationale makes complete sense, but if we follow it through, we would have to rename many other variables:

- `totalForgoneAmount` to `getTotalForgoneAmount`
- `ipfsCID` to `getIpfsCID`
- `shape` to `getShape`
- etc.

So instead of doing that and having to continuously decide "should we add 'get', or not?", I suggest doing away with the decision by not using `get` anywhere in Airdrops repo.

smol-ninja:
Good point about renaming variables. 

[...]

Issue #135, opened 2025-03-27
Rename "TOTAL_PERCENTAGE" to "TRANCHES_TOTAL_PERCENTAGE"


PaulRBerg:
So that it cannot be confused for any vesting unlock percentage

Issue #134, opened 2025-03-27
Refactor `claim` by moving its interface into each Merkle contract


smol-ninja:
### Context

The claim function is implemented within the Merkle contracts. However, the public function and interface are defined in the [`SablierMerkleBase`](https://github.com/sablier-labs/airdrops/blob/staging/src/abstracts/SablierMerkleBase.sol#L107) which led to writing natspecs that apply universally to all Merkle campaigns.

### Problem

While the claim function is implemented differently in each Merkle contract, they all share the same interface and [common natspecs](https://github.com/sablier-labs/airdrops/blob/staging/src/interfaces/ISablierMerkleBase.sol#L81-L100). Ideally, the natspecs should be specific to the individual Merkle contracts rather than shared across all implementations.

### Solution

- [x] Rename `claim` to `_preProcessClaim ` and change its visibility to `internal` in `SablierMerkleBase`
- [x] Remove `claim` function from `ISablierMerkleBase` interface
- [x] Add bespoke `claim` function in each Merkle contract
- [x] Rename `_claim` to `claim` and change its visibility to `external` in each Merkle contract
- [x] Call `_preProcessClaim ` inside `claim`

@sablier-labs/evm I created the issue without initiating a discussion because I think its very obvious. However, please feel free to object if you disagree.

PaulRBerg:
Makes complete sense to me. Big brain idea.

smol-ninja:
Good idea. Yes we can.

smol-ninja:
Yes we can implement `postprocess` but it wont be CEI. It will be CEIE because `MerkleBase` only gas CE where each merkle contract will have CEI..

andreivladbrg:
I can't say I'm strongly in favor of this change, but I'm not strongly against it either. Since it seems you both agree, we will go ahead and implement it.

The reasons I would be against it are:
- [Common NatSpec](https://github.com/sablier-labs/airdrops/blob/dc6bbd693aa57235221fcceae12854b9f389f83c/src/interfaces/ISablierMerkleBase.sol#L81-L100) would be duplicated across all 4 interfaces (and to each new model we will implement).
- There isn't much information added in each claim function (only for VCA can I say it adds meaningful info).

An alternative solution would be to add a `@dev` NatSpec in the [internal function](https://github.com/sablier-labs/airdrops/blob/dc6bbd693aa57235221fcceae12854b9f389f83c/src/SablierMerkleVCA.sol#L184-L185).

---

[...]

PR #150, opened 2025-04-20
refactor: move claim definitions into merkle contracts

smol-ninja:
Small but important refactors before the audit.

## Changelog

- Closes https://github.com/sablier-labs/airdrops/issues/136
- Closes https://github.com/sablier-labs/airdrops/issues/135
- Closes https://github.com/sablier-labs/airdrops/issues/134

### DIFF
--- a/src/SablierMerkleLT.sol
+++ b/src/SablierMerkleLT.sol
@@ -44,7 +44,7 @@ contract SablierMerkleLT is
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @inheritdoc ISablierMerkleLT
-    uint64 public immutable override TOTAL_PERCENTAGE;
+    uint64 public immutable override TRANCHES_TOTAL_PERCENTAGE;
 
     /// @inheritdoc ISablierMerkleLT
     uint40 public immutable override VESTING_START_TIME;
@@ -87,7 +87,7 @@ contract SablierMerkleLT is
             totalPercentage += percentage;
             _tranchesWithPercentages.push(params.tranchesWithPercentages[i]);
         }
-        TOTAL_PERCENTAGE = totalPercentage;
+        TRANCHES_TOTAL_PERCENTAGE = totalPercentage;
     }
 
     /*//////////////////////////////////////////////////////////////////////////
@@ -106,8 +106,8 @@ contract SablierMerkleLT is
     /// @inheritdoc SablierMerkleBase
     function _claim(uint256 index, address recipient, uint128 amount) internal override {
         // Check: the sum of percentages equals 100%.
-        if (TOTAL_PERCENTAGE != uUNIT) {
-            revert Errors.SablierMerkleLT_TotalPercentageNotOneHundred(TOTAL_PERCENTAGE);
+        if (TRANCHES_TOTAL_PERCENTAGE != uUNIT) {
+            revert Errors.SablierMerkleLT_TotalPercentageNotOneHundred(TRANCHES_TOTAL_PERCENTAGE);
         }
 
         // Calculate the tranches based on the unlock percentages.
--- a/src/SablierMerkleLT.sol
+++ b/src/SablierMerkleLT.sol
@@ -7,7 +7,6 @@ import { uUNIT } from "@prb/math/src/UD2x18.sol";
 import { UD60x18, ud60x18 } from "@prb/math/src/UD60x18.sol";
 import { Lockup, LockupTranched } from "@sablier/lockup/src/types/DataTypes.sol";
 
-import { SablierMerkleBase } from "./abstracts/SablierMerkleBase.sol";
 import { SablierMerkleLockup } from "./abstracts/SablierMerkleLockup.sol";
 import { ISablierMerkleLT } from "./interfaces/ISablierMerkleLT.sol";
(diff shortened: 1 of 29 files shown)


src/SablierMerkleLT.sol
```diff
@@ -171,17 +182,17 @@ contract SablierMerkleLT is
         }
 
         // Load the tranches in memory (to save gas).
-        MerkleLT.TrancheWithPercentage[] memory tranchesWithPercentages = _tranchesWithPercentages;
+        MerkleLT.TrancheWithPercentage[] memory tranchesWithPct = _tranchesWithPercentages;
```
andreivladbrg, line 185: what is the motivation for this renaming?

IMO it's inconsistent — it would be the only place where the short form is used  
we can have these rules:

#### Full word
- type declarations like `DataTypes.sol`
- state variable declarations
- function names
- error names

#### Short form
- local function variables (like this one)
- function params, when the type itself contains the full word

wdyt? *(hope i haven't missed anything)*

src/interfaces/ISablierMerkleLL.sol
```diff
@@ -29,4 +29,26 @@ interface ISablierMerkleLL is ISablierMerkleLockup {
...
+    ///
+    /// @dev It emits a {Claim} event. If the vesting end time is in the future, it creates a Lockup Linear stream,
+    /// otherwise it transfers the tokens directly to the recipient.
+    ///
+    /// Requirements:
+    /// - The campaign must not have expired.
+    /// - `msg.value` must not be less than the value returned by {calculateMinFeeWei}.
+    /// - The `index` must not be claimed already.
+    /// - The Merkle proof must be valid.
+    /// - The requirements for {SABLIER_LOCKUP.createWithTimestampsLL} must be met.
```
andreivladbrg, line 47: i would be better to say from `{ISablierLockup.createWithTimestampsLL}` - they are placed under the interface (also "from" instead of "for")

src/interfaces/ISablierMerkleLL.sol
```diff
@@ -29,4 +29,26 @@ interface ISablierMerkleLL is ISablierMerkleLockup {
...
     function VESTING_TOTAL_DURATION() external view returns (uint40);
+
+    /*//////////////////////////////////////////////////////////////////////////
+                               NON-CONSTANT FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @notice Makes the claim.
+    ///
+    /// @dev It emits a {Claim} event. If the vesting end time is in the future, it creates a Lockup Linear stream,
+    /// otherwise it transfers the tokens directly to the recipient.
```
andreivladbrg, line 40: this explanation should be placed under `notice` natspec not `dev` - applies to all claim functions

src/interfaces/ISablierMerkleVCA.sol
```diff
@@ -47,4 +47,32 @@ interface ISablierMerkleVCA is ISablierMerkleBase {
...
     /// @notice Retrieves the total amount of tokens forgone by early claimers.
     function totalForgoneAmount() external view returns (uint256);
+
+    /*//////////////////////////////////////////////////////////////////////////
+                               NON-CONSTANT FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @notice Makes the claim.
+    ///
+    /// @dev It emits a {Claim} event and transfers the tokens directly to the recipient.
```
andreivladbrg, line 57: it must mention the full amount or calculated amount

andreivladbrg (commented):
great work, left some comments:

src/interfaces/ISablierMerkleLL.sol
andreivladbrg: 9b6a33e09f13eca6460aee236855cb71d1a3eb26

src/interfaces/ISablierMerkleLL.sol
andreivladbrg: 9b6a33e09f13eca6460aee236855cb71d1a3eb26

src/SablierMerkleLT.sol
smol-ninja: I think as long as they are internal variables and not confusing, it is fine by me.

Nevertheless, what variable name should I use for `tranchesWithPct`? We cannot use `tranchesWithPercentages` or `_tranchesWithPercentages`, though we can use `tranchesWithPercentages_` but imo having too many `_` causes confusion as one might confuse between `_tranchesWithPercentages` and `tranchesWithPercentages_` in the same code scope.

andreivladbrg: fair enough

smol-ninja:
Thanks for the review @andreivladbrg. If there are no more suggestion, can you please approve the PR?

andreivladbrg (approved):
shogun_lgtm

smol-ninja:
I will be using "merge commit" so that it creates 3 commits for 3 separate issues.

andreivladbrg:
> I will be using "merge commit" so that it creates 3 commits for 3 separate issues.

alright. they are small changes, so it's fine to squash IMO in these cases

smol-ninja:
Small but could be breaking due to the change in the function name.

Issue #164, opened 2025-06-11
Add "IS_SABLIER_MERKLE" boolean constant in


PaulRBerg:
Context here: https://github.com/sablier-labs/indexers/issues/130#issuecomment-2949190592

The task is to implement a simple `IS_SABLIER_MERKLE` constant set to `true` in the [`SablierMerkleBase`](https://github.com/sablier-labs/airdrops/blob/8cbf863b74339e0386231cd16c815d21f62efc0f/src/abstracts/SablierMerkleBase.sol) contract.

@sablier-labs/evm Can we implement this ASAP this week? It should be easy-peasy.

smol-ninja:
Sure.

PR #165, opened 2025-06-11
feat: adds `IS_SABLIER_MERKLE` constant

smol-ninja:
Closes https://github.com/sablier-labs/airdrops/issues/164


andreivladbrg (commented):
shouldn't it be placed only in the `SablierMerkleLockup` instead?

andreivladbrg (approved):

PR #170, opened 2025-06-23
chore: remove smt

PaulRBerg:


smol-ninja (changes requested):
- We should remove [smtchecker.yml](https://github.com/sablier-labs/airdrops/blob/chore/remove-smt/.github/workflows/run-smtchecker.yml) file as well.
- We should rename `bnb` to `bsc` in `rpc endpoint` section in `foundry.toml`.

smol-ninja (approved):

PR #176, opened 2025-07-11
feat: dynamic domain separator

smol-ninja:
Addresses [finding 3](https://cantina.xyz/code/ea8ed3af-8447-4295-83d4-f2cdc19b1c10/findings?finding=3)

### DIFF
--- a/src/interfaces/ISablierMerkleBase.sol
+++ b/src/interfaces/ISablierMerkleBase.sol
@@ -28,10 +28,6 @@ interface ISablierMerkleBase is IAdminable {
     /// @notice Retrieves the address of the comptroller contract.
     function COMPTROLLER() external view returns (address);
 
-    /// @notice The domain separator, as required by EIP-712 and EIP-1271, used for signing claim to prevent replay
-    /// attacks across different campaigns.
-    function DOMAIN_SEPARATOR() external view returns (bytes32);
-
     /// @notice The cut-off point for the campaign, as a Unix timestamp. A value of zero means there is no expiration.
     /// @dev This is an immutable state variable.
     function EXPIRATION() external view returns (uint40);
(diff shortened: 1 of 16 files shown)


src/interfaces/ISablierMerkleBase.sol
```diff
@@ -54,6 +50,10 @@ interface ISablierMerkleBase is IAdminable {
     /// @notice Retrieves the name of the campaign.
     function campaignName() external view returns (string memory);
 
+    /// @notice The domain separator, as required by EIP-712 and EIP-1271, used for signing claim to prevent replay
+    /// attacks across different campaigns.
+    function domainSeparator() external view returns (bytes32);
```
andreivladbrg, line 55: should we add more notes regarding the motivation behind the cache? (the change of chain id)

tests/utils/Utilities.sol
```diff
@@ -5,7 +5,7 @@ import { MessageHashUtils } from "@openzeppelin/contracts/utils/cryptography/Mes
 import { StdConstants } from "forge-std/src/StdConstants.sol";
 import { SignatureHash } from "src/libraries/SignatureHash.sol";
 
-library Utilities {
+abstract contract Utilities {
```
andreivladbrg, line 8: why a contract and not a lib? there are now fewer places where we use it

andreivladbrg (commented):
generally looks good 👍

src/interfaces/ISablierMerkleBase.sol
smol-ninja: Its added in the main contract.

https://github.com/sablier-labs/airdrops/blob/fec54e4697120ef5ec95f414b4d5f0864f782765/src/abstracts/SablierMerkleBase.sol#L30-L32

andreivladbrg: right

tests/utils/Utilities.sol
andreivladbrg: i think it makes more sense for this to be a library, and maybe move it under `src/tests/`?

what do you think about renaming it to `SignatureUtils`?

smol-ninja: I dont think we need it in `src/tests` as its only used in this repo. Happy to call it library though. Also, calling it `Utilities` would mean it can be used to add more functions in it. I think the name is fine.

smol-ninja:
Can you plz approve the PR if looks good now?

PR #179, opened 2025-07-23
docs: update license ens

PaulRBerg:


smol-ninja:
> we have to get hasty with that monorepofication

If you want, I can do it next month.

smol-ninja (approved):
LGTM

